PR #12, opened 2018-08-12
Introducing remote API

Bennedetto:
Introducing remote API to offer a possibility to get generated map images (png) from the robot.
This is to be used in 3rd party apps like OpenHab,..


jorsmm:
@Bennedetto, you can use 
```
image.scale(scale,Jimp.RESIZE_NEAREST_NEIGHBOR);
```
after
```
image.crop( xMin-border, yMin-border, xMax-xMin+2*border, yMax-yMin+2*border );
``` 
instead of 
```
for (var xOffset=0; xOffset<scale; xOffset++) {
	for (var yOffset=0; yOffset<scale; yOffset++) {
```
It is way faster than iterating (and you can use bigger scales, like x4 x8 without afecting performance).

Apart from that, I think you could reuse colors instead of generating for each pixel.

This could be the code:

```
 this.app.get("/api/remote/map", function(req,res){
        WebServer.FIND_LATEST_MAP(function(err, data){
            if(!err) {
                var scale = 4; //TODO: define using post/get params
                var border=2; //secure border on all sides!
                var width=1024;
                var height=width;
                //create current map
                new Jimp(width, height, function(err, image) {
                    if(!err) {
                        //for cropping
                        var xMin=width;
                        var xMax=0;
                        var yMin=height;
                        var yMax=0;
                        //variables for colors
                        let color;
                        let color1=Jimp.rgbaToInt(102,153,255,255);
                        let color2=Jimp.rgbaToInt(0,118,255,255);
                        //let color3=Jimp.rgbaToInt(0,118,255,255);
                        let color3=Jimp.rgbaToInt(82,174,255,255);
                        data.map.forEach(function (px) {
                            //calculate positions of pixel number
                            var yPos = Math.floor(px[0] / (height*4));
                            var xPos = ((px[0] - yPos*(width*4))/4);
                            //cropping
                            if (yPos>yMax) yMax=yPos;
                            else if (yPos<yMin) yMin=yPos;
                            if (xPos>xMax) xMax=xPos;
                            else if (xPos<xMin) xMin=xPos;
                            if (px[1] === 0 && px[2] === 0 && px[3] === 0) {
                                color=color1;
                            }
                            else if (px[1] === 255 && px[2] === 255 && px[3] === 255) {
                                color=color2;
                            }
                            else {
                                color=color3;
                            }
                            //set pixel on position
                            image.setPixelColor(color, xPos, yPos );
                        });
                        //crop to the map content
                        image.crop( xMin-border, yMin-border, xMax-xMin+2*border, yMax-yMin+2*border );  
                        image.scale(scale,Jimp.RESIZE_NEAREST_NEIGHBOR);
                        //use process.env.VAC_TMP_PATH to define a path on your dev machine - like C:/Windows/Temp
                        var tmpDir = (process.env.VAC_TMP_PATH ? process.env.VAC_TMP_PATH : "/tmp");
                        var directory = "/maps/";
                        if (!fs.existsSync(tmpDir + directory)){
                                fs.mkdirSync(tmpDir + directory);
                        }
                        //delete old image (keep the last 5 images generated)
                        var numerOfFiles =0;
                        fs.readdirSync(tmpDir + directory)
                        .sort(function(a, b) {return a < b;})
                        .forEach(function (file) {
                            numerOfFiles++;
                            if (numerOfFiles>5)  {
                                fs.unlink(tmpDir + directory + file, err => { if (err) console.log(err) });
                                //console.log( "removing " + toClientDir + directory + file);
                            }
                        });
                        //Position on the bitmap (0/0 is bottom-left)
                        var homeX=0;
                        var homeY=0;
                        var imagePath;
                        //save image
                        var date = new Date();
                        var dd = (date.getDate() < 10 ? '0' : '') + date.getDate();
                        var mm = ((date.getMonth() + 1) < 10 ? '0' : '') + (date.getMonth() + 1);
                        var yyyy = date.getFullYear();
                        var HH = (date.getHours()<10?'0':'') + date.getHours();
                        var MM = (date.getMinutes()<10?'0':'') + date.getMinutes();
                        var SS = (date.getSeconds()<10?'0':'') + date.getSeconds();
                        var fileName = yyyy + "-" + mm + "-" + dd + "_" + HH + "-" + MM + "-" + SS + ".png";
                        imagePath = directory + fileName;
                        image.write(tmpDir + imagePath);
                        //set charger position
                        homeX = ((width / 2) - (xMin - border)); //not verified yet
                        homeY = ((height / 2) - (yMin - border)); //not verified yet
                        //define return value
                        res.json({
                            mapsrc : imagePath,
                            charger : [ homeX, homeY ]
                        });
                    } else {
                        res.status(500).send(err.toString());
                    }
                });
            } else {
                res.status(500).send(err.toString());
            }
        });
    });
```

jorsmm:
@Bennedetto could you please add another method to the api (something like `/api/remote/map_and_path`) similar to `/api/remote/map` but including the path and the icons of the charger/robot, painted on the image?
It could be very useful for example for a Telegram Bot to control the Robot and show where it is.
Thanks in advance

Bennedetto:
@jorsmm thanks a lot for that hints!! I will implement and push those in the upcomming days!

I already planned to make it adjustable in order to request images with:
- only the map
- only the path
- map and path
- draw current robot
- draw charger
- in a custom scale
I just need a good way to provide and evaluate those parameters. Since I am new to node.js I  appreciate getting possible ways to go here. But in general I would like to avoid introducing too many urls - I prefer to keep as less urls as feasible and make them parametrizeable instead.

Bennedetto:
One bug remaining;
If robot or charger should be drawn, the map is not generated successfully.
Since I am developing on windows (it worked there) and I use atm the github repository to sync my rpi with the current code in order to deploy - this pr is atm not bugfree.
To be fixed tomorrow.

PR #13, opened 2018-08-19
Introducing remote api

Bennedetto:
Added remote api to support 3rd party instances in getting generated maps as images from the vacuum


Bennedetto:
Here we go: You might use this binary:
https://github.com/Bennedetto/Valetudo/releases/tag/0.0.4.a

krabbs:
Working great! Just perfect. I don't declare following as an issue because I'm not sure yet. however i am not sure if the tmpdir including the requested picture will be ever cleared. In fact openhab or any other client will bloat the memory space?

Bennedetto:
Valetudo will only keep the last generated 6 maps. As a new map is generated, the oldest map will be deleted. This should prevent ending up with a blown memory space.

Hypfer:
While I really appreciate the efforts, I'm unsure if it's a good idea to have two different ways of providing the map (first being the current canvas implementation).

The Live map could be changed to use this png map as well, although I'm not sure if thats a good idea either. It would reduce cpu usage and speed up the live map on mobile devices, though. 🤔

Hypfer:
Since #15 will introduce a config file in /etc/valetudo, I suppose this code should store its maps at the same location.

Bennedetto:
I suggest to keep the `/tmp/` folder for the generated maps, since they are only "buffered" and furthermore they are to be removed after a reboot. But if you want the maps to be generated to `/etc/valetudo` anyway, I can of course point the saving/loading to this directory.

Regarding the second approach of generating maps;
The remote api offers a simple to use method to get the most recent map as an (configurable) image + all required positions / coordinates to be used by third party apps like smart home systems or apps. Thus it allows valetudo users to completely abandon the cloud service - since using it does not provide any additional benefit anymore.
Moreover using the generated image does - as you already pointed out - provide a speed benefit in valetudo: As soon as the generated image is used instead of outsourcing the generation to the user. While I implemented the generation of the image, I changed the current map view to a version loading the image; This observations had been made (using my i3 desktop machine):

_Generating the image in to browser:_
loading json image: ~180ms
time until map is shown: ~3.5 sec
*lack in browser: ~3.5 sec*

_Generate and load image  from the robot:_
generating png: ~500ms (on my i3 desktop)
loading image: ~21ms
time until map is shown: ~550ms
*lack in browser: ~0ms* (as png is generated and loaded async)

That said, the general oberservationis: The smaller the map gets, the better the current approach is. The bigger the map gets, the better the png-map approach gets. The point of those lines to cross reached quite early on its cleaning route.

I have not yet uploaded the modified map as I am not fully satisfied with its working state atm (scaling is some kind of buggy in the frontend). And I wanted to wait for this PR to go through.

Generally spoken, at this very moment I see the second approach of generating a map as an addon to provide more functionality for further and easy integration of the roborock into smarthomes without neglecting users without smarthome-requirement.

Hypfer:
Fair point. Okay, I'll merge this.

Thank you for your contribution

PR #14, opened 2018-08-21
Add sound volume setting

maxpautsch:
@Hypfer: Thank you for this great software!
I added the parameter 'sound volume' to the configuration tab. Would it be better to remove the save button and configure the new value every time the slider is moved?

settings_volume
settings

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -264,16 +264,20 @@
             <hr style="width:98%; opacity: 0.3">
 
             <section id="robot-control-buttons">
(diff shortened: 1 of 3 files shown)


maxpautsch:
this adds also a 'test sound' button: https://github.com/maxpautsch/Valetudo/commit/2fd2deade2714913ab764b42d4e5bf98bd962760

Hypfer:
Please combine your changes into less commits and send a new PR after everything is finished

PR #15, opened 2018-08-24
Add zoned cleanup, go to, spot clean and sound volume configuration

maxpautsch:
I added the following functionalities:
- zoned cleanup
- go to
- sound volume configuration
- spot clean
In order to use zoned cleanup and the go to functionality, i introduced a config file. See README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,21 @@ Valetudo does not feature access controls and I'm not planning on adding it sinc
 You could just put a reverse proxy with authentication in front of it if you really need it.
 
 Please don't just forward the port to make it accessible on the go..
+
+### Zoned Cleanup and Go To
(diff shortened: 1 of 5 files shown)


maxpautsch:
The defined zones and go to locations can be selected at the home screen home

Hypfer:
Please move the config.json to /etc/valetudo/config.json and add code which creates it if it doesn't already exist. /usr/local/bin is no place for a config file
Also, a comment explaining Vacuum.prototype.startCleaningZone would be nice since I have no idea why that for loop exists.

Everything else looks fine. Thank you for your contribution

maxpautsch:
Tested on target and it works.

Hypfer:
The goto feature is incredibly useful for servicing the robot since I've placed the dock right under the couch.

Thanks 👍

PR #19, opened 2018-09-01
Rework of Map-View

Bennedetto:
This is the initial commit of the "new" map view.
It now generates the map on the robot and transfers it as a png image. There is still some work to do before it looks as nice as the previous version. But as the image scales (not implemented yet) automatically to the available screen size, it should look less pixelated.
Furthermore it prevents lacks in the browser as the map grows (since it is no longer generated in the browser).
This version also introduces the drag-and-drop goto method in the map view.


WolfspiritM:
Hi.
I now switched to Valetudo aswell but don't like the current map view at all.
I prefer the way the app handles view and zoom.
I don't see any performance issue generating the image in the frontend as long as it's just done once every few seconds.

A few days ago I've been working on a map that scales via pinch/mousewheel and is dragable.
It's still a bit buggy but that's what I came up with (missing bot and route for now and is still a bit "buggy") I don't have any issue with pixelation:

https://i.gyazo.com/0a6af3bc713123332f07061ad4c4dfc3.gif

Another reason why I think doing that stuff (beside the actual image) in the frontend is that markers and paths shouldn't zoom with the image.

maxpautsch:
@WolfspiritM: do you plan to publish these changes? They seem pretty cool!

Bennedetto:
I would also like to see the performance optimized code. Since the generation of the image leads to a 3 seconds lack all 5 seconds (so 5 seconds usable, 3 seconds non usable, 5 seconds usable,..), this approach seemed not feasible for me (thats why I changed to the onboard generated map).

Hypfer:
I think I will reject this PR since @WolfspiritM is working on his rewrite of the client

leonardpitzu:
@WolfspiritM is your branch buildable? i get a very long series of errors that the --no-bytecode option failed... i am sure that i must have some wrong setup but can't figure it out. I'm on Ubuntu + latest LTS nodejs/npm.

PR #30, opened 2018-10-03
Add a general info section with firmware version

TobiasSchaffner:
Add a general info tab for general information like type, firmware version and build to the settings page. Add firmware and build version to that tab.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -1635,6 +1635,10 @@
 
     <template id="settings.html">
         <ons-page id="settings-page">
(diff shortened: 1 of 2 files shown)


Hypfer:
Why are you using cat to read the file contents?
Please use
https://nodejs.org/api/fs.html#fs_fs_readfile_path_options_callback

Hypfer:
Also, there's no point in doing this synchronous

Issue #40, opened 2018-11-03
Error 500 when using map function while charging


ovaltineo:
Valetudo worked properly on older firmware but now returns error 500 when using map function on Gen 1 firmware v11_003514 (downloaded via FloleVac app).

Crazyachmed:
Works fine for me on 3514. I had this issue only before I let the robot run for the first time.

ovaltineo:
It definitely has a map of the house - I can see the map in the Xiaomi app and FloleVac app.  It was giving the error 500 when I tried it while charging.  I tried now while it is doing a zoned cleaning and it worked.

ovaltineo:
Sure it enough, it returns error 500 when it went back to dock to charge.  Is this a known bug?

Crazyachmed:
Now that you are saying it, it displays an old map as soon as it hits the dock

Mdleal:
I am seeing the issue with a Gen 2 running v11_001633

herrwusel:
@Mdleal same here: Rockrobo S50 with 1633 firmware.

moskovskiy82:
Same with me 1780

JohnRev:
I have the same issue with 1644 and above.
The issue does not occur on 1518.

I have noticed that in the newer firmwares, the map file in `/mnt/data/rockrobo/rrlog/` gets renamed from `navmap344327.ppm.0000.gz` to `navmap344327.ppm.0000.rr.gz`, and so the regex on https://github.com/Hypfer/Valetudo/blob/2b38868bf5df113e6fe66788502c9822ef68b71c/webserver/WebServer.js#L925 and https://github.com/Hypfer/Valetudo/blob/2b38868bf5df113e6fe66788502c9822ef68b71c/webserver/WebServer.js#L928 will fail.

I tried modifying the regex to something like:
`/^navmap([0-9]+)\.ppm\.([0-9]{4})(\.rr)?\.gz$`.
With this updated line, the navmap file is correctly identified and located, but the decryption fails. I am not sure how those `.rr.gz` files are encrypted.

Hypfer:
Firmware 1518 is the last firmware that uses the old map format IIRC.
It could be possible that the .rr.gz navmap is in a completely different format which is not (yet) supported.

A sample .rr.gz Mapfile should be interesting. If you feel like sharing your map please attach it to #44

leonardpitzu:
i'm on 1633 - what do you need, the nvmap archive or something more?

[...]

PR #59, opened 2018-12-19
Fix directory scanning

JohnRev:
This fixes the error described in https://github.com/Hypfer/Valetudo/issues/40#issuecomment-446504197

### DIFF
--- a/webserver/WebServer.js
+++ b/webserver/WebServer.js
@@ -913,7 +913,7 @@ WebServer.FIND_LATEST_MAP_IN_ARCHIVE = function(callback) {
             let logFileName;
 
             for(let i in folders) {
-                const folder = folders.pop();
+                const folder = folders[i];
                 try {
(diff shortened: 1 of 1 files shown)


JohnRev:
@Hypfer: I may have inadvertently introduced a bug in this pull request (or at least discovered an existing bug ... however you wanna put it).
This line: https://github.com/Hypfer/Valetudo/blob/d74f081a047b0c6c6fbed42cf02785011500b1ec/webserver/WebServer.js#L914 sorts `folders` in a descending manner. To scan for the latest folder, it makes sense to scan in an ascending manner, so it should instead read as `folders = folders.sort().reverse();` .
Same thing with line https://github.com/Hypfer/Valetudo/blob/d74f081a047b0c6c6fbed42cf02785011500b1ec/webserver/WebServer.js#L938

The folder sorting definitely needs to be reverse. I am unsure about the map files sorting though.

PR #61, opened 2018-12-21
Scan newest folder first

JohnRev:
Following my pull request https://github.com/Hypfer/Valetudo/pull/59, I noticed a new bug arising where an older map would be shown. 
This is because ` folders = folders.sort(); ` sorts the directory names in ascending order. However, the newest folder would be found if sorted by descending order.

I am unsure if the same would be needed for `possibleMapFileNames = possibleMapFileNames.sort();`, so I didn't push this change yet.


cryptomilk:
I've tested this and it worked for me!

PR #62, opened 2018-12-21
Update filename regex and update README.md with patched rrlogd info

JohnRev:


cryptomilk:
Tested and works for me!

PR #63, opened 2018-12-21
Change config file location to /mnt/data

cryptomilk:
If you flash a new firmware, the config in /etc/valetudo will be gone.
We should use the persitant flash memory for that.


metronidazole:
To migrate to this, can we just shutdown valetudo, move the files, upload the new build and start valetudo? Or does anything else need to be modified?

cryptomilk:
You can just copy the directory with the config from /etc to /mnt/data when the service is not running.

PR #78, opened 2019-01-09
use relative path to api to allow hosting with subpaths

matthiasharrer:
Hi,

I think this should do for accessing the map using a reverse proxy setup until the map is either included in the main frontend or removed :)


Hypfer:
image
Why is the build failing? 🤔

PR #87, opened 2019-02-10
Added Valetudo release number to Settings>Info

Bennedetto:
This commit enables user to determine, which valetudo version is installed. Therefore, the "Settings>Info" got extended by "Valetudo version". This number is automatically read out of the "package.json", so no additional maintenance required for this feature.

image


Hypfer:
Nice work

Issue #42, opened 2018-11-12
Timezones


stefanct:
The roborock firmware uses its own timezone configuration: even if you configured the timezone of the ubuntu system correctly and `date` shows the right time the firmware still uses its own tz.
The timers of the roborock (stored in `/mnt/data/rockrobo/apptimer.cfg`) that are also set by valetudo's timer configuration dialog abide this tz and will happily wake you in the middle of the night (been there, done that).

To fix this you have to set two files in `/mnt/data/rockrobo` correctly.
For example, to set the timezone to `Europe/Vienna` you can execute this via ssh:

    echo 'Europe/Vienna' > /mnt/data/rockrobo/timezone
    ln -sf /usr/share/zoneinfo/Europe/Vienna /mnt/data/rockrobo/localtime

Alternatively, `mirobo` does also the correct thing (although wasting a few bytes):
    mirobo  --ip ... --token ...  timezone Europe/Vienna

It would be very nice to be able to do that within valetudo or at least display the timezone information on the timer page ;)

LemarinelNet:
Hi
I'm also a victim of the sudden-in-need-of-cleaning-in-middle-of-the-night robot !
Set the timer last week, and this morning, the robot woke up at 6h30, instead of 13h30...
I've now setup timezone info, but the "date" command still return the wrong time ? Is a reboot needed ?

stefanct:
The system settings are distinct from those used by the xiaomi software. To set the timezone of the system you can use `dpkg-reconfigure tzdata` as on ordinary Ubuntu machines. I presume you did not change that yet? Cf. `cat /etc/timezone`.

stefanct:
Well, that's from dustcloud not Valetudo... so yes, if you know about the issue before changing the firmware that's the easier way to do it but being able to change it later makes sense nevertheless IMHO.

PR #93, opened 2019-02-16
Improved timer ui

Bennedetto:
Changed timer ui and added some features:
- UI does now seperate between cleaning and dnd timer
- UI does now allow easy configuration of cleaning timer
- UI does now allow configuration of timezone (fixes https://github.com/Hypfer/Valetudo/issues/42)

image

image

image

Tested on Gen2 robots - I do not have any possibility to test it on Gen1 bots, so if *you* have the possibility to test, please do so and let me know if it worked properly.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -2200,84 +2200,120 @@
     <template id="settings-timers.html">
         <ons-page id="settings-timers-page">
                 <ons-dialog id="add-timer-dialog">
+                    <div style="text-align: left; padding: 10px;">
+                        <form id="add-timer-form">
+                            <!--Date-->
+                            <p>Select the date the timer should be triggered on.<br/><i>Empty means any month/day.</i></p>
+                            <ons-row>
+                                <ons-col width="100px" align="center">Month</ons-col>
+                                <ons-col><ons-input type="number" placeholder="*" min="1" max="12" name="month"></ons-input></ons-col>
+                            </ons-row>
+                            <ons-row>
+                                <ons-col width="100px" align="center">Day</ons-col>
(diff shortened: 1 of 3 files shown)


stefanct:
Good work @Bennedetto! One thing I still miss is a way to edit existing timers. With these changes it is still only possible to disable or delete existing timers and not modify them.

CodeFinder2:
Hi, I tried this on my gen1 with valetudo v0.4.0 and it's not working:

https://ibb.co/hV5cVP8
https://ibb.co/SyxjtyZ

Any other way to do this? Can I fix this to make it working on my gen1?

stefanct:
Did you reboot the system after the change? I am not 100% sure it is needed but would not be surprised at all.

CodeFinder2:
Just rebooted, no changes... 😕

CodeFinder2:
Okay, wait might got something wrong: I assumed changing the timezone also changes the Setting > Info > Timezone data. Is that correct? If I change the timezone there, it stays changed in that dropdown box but it is not changed under Settings > Info.... 🙈 

I just followed [this tutorial](https://github.com/dgiese/dustcloud/wiki/Vacuum-Robots-CCC-to-CE-conversion) to make my CCC gen1 a CE gen1 and now it correctly says "Europe/Berlin" under Settings > Info. However, timezone settings in "Timers" are (still) not reflected in Settings > Info. Is that intended?

But can confirm for my gen1 that changing the timezone under "Timers" changes the timezone (only) there correctly (also persistent after a reboot). 😊👍

PR #94, opened 2019-02-16
Integrate improved /zone map into main UI

matthiasharrer:
This integrates the map from /zone into the main UI
It now also has additional features:
- Allows to add multiple zones for zoned cleanup
- Allows to delete/move/resize these zones

Tapping onto the map sets a gotopoint which can be executed by clicking the gps symbol.
For zoned cleanup press the + button to add a zone move / resize it and press play to start the cleanup.

The drawing logic is still in separate files to allow moving it to a different project later on.


cryptomilk:
I think you need to display a help about this somewhere. Like add a ? button to open a popup describing how it works.

cryptomilk:
The interface works fine with Firefox on the desktop, but it doesn't work in Firefox on Android. I'm not able to select a goto point, I'm not able to zoom with two fingers.

Nebukadneza:
Same here — Chromium and `Browser` on Android do work, though. Pity for Firefox … :(.

Still, awesome work, I really like the new and now fully integrated map!

matthiasharrer:
I guess I used an API which is not yet available on Firefox Mobile (https://caniuse.com/#search=pointer%20events) and also not on Safari on iOS.

I will try to implement it using the mouse and touch events separately as soon as I find the time.
Concerning the help text .. maybe someone else could try to add a ? button which opens an overlay containing the appropriate descriptions?

matthiasharrer:
It should now work in Firefox Mobile / Safari although performance is still a lot better in Chrome/Chromium.

cryptomilk:
Awesome, works just fine! You should rebase your branch on master :-)

matthiasharrer:
Awesome how many pull requests were merged since I opened this PR!
I rebased it and did a short test using my v2 robot and it seems to work as intended :)

Nebukadneza:
Ack, works perfectly fine in firefox, thanks a LOT!

Space2Man:
Wow, this looks great! Just installed v0.2.0 and things start to get even cooler then expected!

Is it possible to add the zones defined in the Map view somehow to the Zones view? So you could add new zones visually and then rename / use them via Zones view?

matthiasharrer:
Its not yet possible, but that is the plan i was working towards. :)

Space2Man:
Perfect!

PR #104, opened 2019-02-23
Adds rudimentary display of zone coordinates to /zone again

matthiasharrer:
This "feature" seems to be missed by some people.
I added it again to the /zone map so people can use it until specifying zones using the map is possible.

This does not change the map integrated in the main UI.

### DIFF
--- a/client/zone/index.html
+++ b/client/zone/index.html
@@ -28,8 +28,9 @@
         body {
             display: grid;
             grid-template-columns: 1fr 1fr 1fr 1fr;
-            grid-template-rows: 1fr 100px;
+            grid-template-rows: 100px 1fr 100px;
             grid-template-areas:
+                "zone zone zone zone"
                 "map map map map"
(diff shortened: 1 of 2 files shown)


imbaimbaimba:
Sorry if this is a dumb question, but how can I update with this feature now that it is implemented? I'd really like to use it, as I need to fix my cords. Till now, I just downloaded a binary from the /releases page and dropped to the robot with WinSCP. I don't know how to compile and stuff.

If this is too dumb of a question, just ignore it :)

matthiasharrer:
Have a look at https://github.com/Hypfer/Valetudo/blob/master/develop/HOWTO.md

I think you can skip 4-6 and you might need to get git / npm from somewhere else than apt-get if your using windows.

imbaimbaimba:
Hmm it doesn't work unfortunately, it just gives me the folder structure, as I can also download it. The (binary I believe), file that I need, I cannot figure out how to get.

Issue #64, opened 2018-12-23
[RELEASE] Provide tarball or zip with binary including deployment dir


cryptomilk:
Hi,

could you please provide a zip or tarball for releases which contain the binary + the deployment directory. Like in the soruce directory when you compile it:

```
+ valetudo
+ deployment
 \ - Readme.md
   - valetudo.conf
```

This way I could add a --valetudo-path to the imagebulider.sh of dustcloud and directly add it when building the firmware.

Thanks.

Hypfer:
Maybe it would be better to provide a .deb package for easy installation/upgrades 🤔

cryptomilk:
Would also be possible :-)

cryptomilk:
But then we should have all together, dummycloud and valetudo.

Hypfer:
stale

PR #105, opened 2019-02-24
add yarn.lock

petabyteboy:
Makes it possible to create reproducible builds of Valetudo.


Hypfer:
While I appreciate the effords, I don't like to throw more dependencies at the project.

petabyteboy:
You mean yarn? I would argue it's not really a dependency, since npm will continue to work fine. Also, have you looked at the dependency graph of npm? yarn has much less sub-dependencies than npm.

Anyway, npm also has a package-lock.json file, so if you use npm you could commit that.

I actually use the yarn.lock file to package Valetudo with yarn2nix to automatically generate firmware images with a freshly compiled Valetudo version, which is not possible in any other way. 
I will keep maintaining a fork with a yarn.lock file to make this possible.

Hypfer:
When `package-log.json` was first introduced, I happily commited it to every repository I was working on and it instantly broke everything in some weird ways which no one could understand or debug. The only solution was to delete that file.
If you dig deeper and ignore every "Just commit it because we've told you you should do that", you'll find out that the reason for this lockfile is to prevent random and unexpected breakage which is exactly what it caused.
I don't know if the situation is besser now or with yarn but also I just don't want to know. It works very well like it is now and there's not really any reason to break that.


Considering the firmware autogeneration:
So.. you use yarn (which is an alternative non-standard package manager) to create nix scripts (which apparently are scripts in some esoteric scripting language) which then do some magic to automatically build a firmware?

That doesn't really sound like anything that should be supported in the standard.
I'm planning to package valetudo into a debian package, though. Thats native to the Ubuntu firmware so it should just work™
See https://github.com/Hypfer/Valetudo/issues/64#issuecomment-449644443

petabyteboy:
Using Nix to generate firmware images works fine for me, and it needs some kind of lockfile to make the build reproducible. I understand that you won't support this, but I will continue to use it to generate my firmwares.

I've never had issues with lockfiles, and the concept makes sense to use tested versions of all dependencies, instead of allowing use of newer versions that could break the build. I've seen projects breaking because of this.
If you have had other experiences I won't try to talk you into committing lockfiles, it's just my opinion that it provides an advantage.

A zipfile with a binary and .conf file would allow me to use precompiled versions in my firmwares, but I think it's really nice that I can compile Valetudo from source easily, so I prefer to do that.

Issue #106, opened 2019-02-24
draw robot and charger on map


petabyteboy:
How can I make Valetudo draw the robot and the charger on the map in the web interface?

Hypfer:
It did that at some point. Guess it's missing in the new 0.2.0 Map.

@matthiasharrer ?

petabyteboy:
I will try to add it back.

matthiasharrer:
You can have a look at https://github.com/matthiasharrer/Valetudo/tree/add-robot-drawing?files=1

It seemed to work but the robots orientation was incorrect. Would be nice if you can figure it out:)

petabyteboy:
But it's exactly the code that was used before..? That's weird.

matthiasharrer:
Yup that's what confused me too.
Did you try it with your robot? Is it working for you?

petabyteboy:
Nope, the angle is very weird.

matthiasharrer:
Ah I think the problem is the flipping of the map. I'm not exactly sure but depending on where the map is read from the api returns a flag that indicates a flipped map. I think we have to account for that with the robots angle too.

petabyteboy:
I mean most of the time it looks okay, but just because of how this code works it will always look a bit weird in some cases.
![foo](https://i.imgur.com/5bLfLgD.png)
![foo](https://i.imgur.com/i2oXuTC.png)
![foo](https://i.imgur.com/3tIrpI0.png)
![foo](https://i.imgur.com/XkKtuHX.png)
You can see that the first two look fine, but the last two look off somehow.
Actually it makes a lot of sense that these angles have not been adjusted for a flipped map.

petabyteboy:
I don't understand this: The whole path should have been flipped to account for the map being flipped. So did the old code somehow flip the y axis, so it is flipped twice now? @Hypfer

petabyteboy:
Nevermind, the way this the path is used the transform is not being applied.

matthiasharrer:
I think the flipping is applied when the meter based path coordinates are transformed to the 1024*1024 px map coordinates. For calculating the angle I used the original path coordinates .. so they miss the flipping.

petabyteboy:
Yes, that's how it is. 
I am having a few other problems where the charger is shown in the wrong place, is that a known issue?

matthiasharrer:
You mean a little bit or completely off? We discussed it in #72 but I never really looked into it.

petabyteboy:
[...]

PR #107, opened 2019-02-24
draw robot and charger (#106)

petabyteboy:


matthiasharrer:
Does this work if the map is not flipped?

petabyteboy:
Probably not, is that a user-changable option?

matthiasharrer:
I'm not sure where it originates, but I think the API returns that flag because in some cases only the map is flipped but not the path.
I don't know in which case this happens though.

petabyteboy:
I could add a new property to the path drawer that saves if the map is flipped or not, as it was set using setFlipped(), and use that to determine the angle correctly in both cases. Does that sound about right?

matthiasharrer:
Yes either that or calculate the angle using already transformed coordinates.

jubudu:
Hello, 
since V0.2.0 neither the robot nor the charger is shown in my map. Downgrading to V0.1.0 gives this nice feature back. 
Is this a bug in V0.2.0?
Kind regards and thanks for this awesome app!

matthiasharrer:
I did not add drawing the charger and the robot when implementing the new map view. However @petabyteboy added it just in this pull request. You can grap the version currently on master as there is no new release so far and will have the robot & charger back.

Hypfer:
image
I think the icons should also be scaled according to the zoomlevel

matthiasharrer:
Although it is just a technical thing that it is not scaled right now, I actually like it because the icons are always clearly visible but don't cover the whole screen and get pixelated when zoomed in..

Hypfer:
image
They do also get pixelated when zoomed in

matthiasharrer:
That's true at some point.. but I don't think zooming into the map this far is interesting. Maybe we just need to limit the scaling :)

PR #108, opened 2019-02-24
prevent crash when drawing robot with empty path

petabyteboy:


matthiasharrer:
I left that check out because I thought I was just experimenting :D

PR #123, opened 2019-03-24
Attributes MQTT topic and Docker build container

westenberg:
+ attributes MQTT topic with statistics and consumables status
+ json_attributes to Home Assistant autodiscovery
+ check for differences with previous MQTT messages, to prevent duplicate messages
+ configurable intervals for state and attributes
+ docker container to create builds on any OS

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM node:8.12.0-alpine
+
+WORKDIR /
+COPY . .
+
+RUN npm install --quiet
+
+ENTRYPOINT [ "npm", "run-script", "build" ]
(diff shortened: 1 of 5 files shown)


Hypfer:
I just woke up because I've realized that this caching feature doesn't make sense.

The way it is now compares pointers which will always be different since its a new image in memory (with most likely exactly the same Pixels)
You need to use .equals and I need better dreams.
How could I miss that before merging.

westenberg:
Do you only mean the image caching or also the caching of state and attributes? If it's only the image caching then it should be changed a bit, so it only creates a new image when needed (coordinate change etc.). I can check it later today.

Hypfer:
It's the same for state and attributes since those are objects and not primitive.
There isn't native equals for objects, though.

Don't waste too much time thinking about that. This polling based approach will most likely be replaced in the foreseeable future anyways.
Also, CPU Cycles on that machine are basically free so there's no real need to invest too much into this problem

westenberg:
I see that I forgot to declare the caching variables (lastStateResult etc.) inside the module. When I add it there then it will compare the current stringified object to the stringified object in e.g. lastStateResult. 

The main goal of this 'cache' check is to reduce the amount of MQTT (duplicate) messages and with that reduce the amount of state updates in Home Assistant.

PR #125, opened 2019-03-25
Bugfix for cache

westenberg:
I tested this cache update fix. It only updates when there are changes in the result. Works even for the map.


westenberg:
It's not merged with the latest master because I wasn't able to run that build on my vacuum.

Hypfer:
This is not required anymore since we don't need to poll when using the cloud interface.

PR #144, opened 2019-04-06
init: Write a valetudo log file

cryptomilk:
Write output of valetudo to a log file.

### DIFF
--- a/deployment/valetudo.conf
+++ b/deployment/valetudo.conf
@@ -4,7 +4,11 @@ description "Valetudo"
 start on filesystem
 stop on runlevel [06]
 
+pre-start script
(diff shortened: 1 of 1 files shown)


Hypfer:
Everything is already logged to `/var/log/upstart/valetudo.log` so I'm unsure if this change is necessary.

cryptomilk:
Oh, I didn't know that.

Hypfer:
Yeah we really need to document that somewhere

cryptomilk:
We need this, as not everything lands in /var/log/upstart/valetudo.log

Hypfer:
Can you give an example?

cryptomilk:
Hmpft, my fault. I did not reset ...

Issue #39, opened 2018-10-30
Memory leak in Valetudo?


dugite-code:
I discovered a small issue when using Home-Assistant to Poll the Map too often (every 10 seconds to get a sort of Live map). The `/tmp` directory filled and the vacuum locked up. A re-boot fixes the issue obviously.

I had done other things on the vacuum in the past so I might have had less space available than normal.

A quick work-around was to throw this into cron
`*/1 * * * * cd /tmp/maps && ls -t /tmp/maps | tail -n +2 | xargs rm --`

dugite-code:
Actually after further looking into the matter it looks like it might be a low memory issue rather than the `/tmp` partition filling up. I'v added a swap file on `/mnt/data` to see if that solves my issues

As you can see Valetudo is taking up 76% of the memory:
![image](https://user-images.githubusercontent.com/12742064/47972887-ae357a00-e0db-11e8-8ad3-486beeadf9c6.png)

Hypfer:
After looking at the [code ](https://github.com/Hypfer/Valetudo/blob/master/webserver/WebServer.js#L403) for 20 minutes, I still have no idea what could be responsible for the memory leak.

However, I've noticed that on every request the robot + charger images are read from flash for no reason so I guess the whole part will be rewritten at some point.
Could you provide a specification for this interface? What endpoints are there? What does Home-Assistant expect? 
I'm not using those home automation frameworks nor did I write said code so I don't have the slightest Idea. 🙃

dugite-code:
Essentially all HA is doing is calling `YOUR.VACUUM.ROBOT.IP/api/remote/map` and then pulling the image using the mapsrc from the json response using a GET request (using the python requests library).

The only real difference between HA and using curl is the speed of the requests, normally HA will do this every 30 seconds I tuned it up to every 10 seconds, I wouldn't think that would cause an issue.

setting  drawRobot and drawCharger to false by calling `YOUR.VACUUM.ROBOT.IP/api/remote/map?drawRobot=false&drawCharger=false&scale=5&border=3&doCropping=true&drawPath=true` looks like it might fix the issue. When drawing the robot and charger at 3min run time it was rising above 25% memory, when not drawing the robot and charger it's bouncing between 16% and 20% at 5min run time.

I'll run this for a while without restarting and let you know if the memory usage stays at ~20% mark

As a side note: Having a further look into the HA settings I've enabled `limit_refetch_to_url_change: True`, This should reduce the fetching of the map image a little.

[...]

PR #150, opened 2019-04-06
build(TravisCI): upgrade Node.js version

timbru31:
Node.js 9 is already EOL. The lts/* version always uses the latest LTS version (currently 10.15.3), the target version from the build command remains unchanged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,14 @@
 language: node_js
 node_js:
-- '9'
+  - lts/*
 dist: trusty
 sudo: true
 
 install:
-- npm install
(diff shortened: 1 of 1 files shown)


Hypfer:
No. The version choice is intentional since there is a memory leak in all nodejs versions newer than 8.

It should be fixed in node 11.10 but honestly I don't want to risk it but rather wait a while until its really stable.
See this Issue for more: https://github.com/Hypfer/Valetudo/issues/39

Hypfer:
Oh it's just the build process. Nvm then

PR #157, opened 2019-04-07
Add ability to edit saved zones on map

matthiasharrer:
Although still work in progress, I wanted the to share the current state of editing saved zones using the map. (Zones-Tab -> Zone configuration -> Modify zones using map)

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -799,11 +799,12 @@
     <template id="zones-configuration-map.html">
             <ons-page id="zones-configuration-map-page">
                 <div id="zones-configuration-container">
-                    <div class="left">
+                    <div class="zones-configuration-header">
                         <ons-back-button>Zones</ons-back-button>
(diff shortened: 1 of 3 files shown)


matthiasharrer:
Current state:
Removed all coordinates based configuration and replaced it by simpler UI.
Goto Configuration is still missing..
Repetition of zones is also still missing 

improved-zone-selection

cryptomilk:
Great work! This looks really promising.

matthiasharrer:
It now also allows to define Goto Points.
It also adds some UI polishing like disabling the save button while the save request is ongoing to prevent multiple requests from happening.

Codequality-wise .. I think it perfectly fits the style of the index.html :)

fenio:
When is this functionality going to be present in official release?
0.3.1 doesn't have anything like that.

PR #159, opened 2019-04-08
Add a Makefile

cryptomilk:
This just adds a simple Makefile:

```
npm install
make
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@
 
 # IntelliJ Idea project files
 /.idea/
+
+# npm package lock file
+/package-lock.json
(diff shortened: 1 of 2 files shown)


cryptomilk:
You should also update deployment/README.md

matthiasharrer:
What's the difference to "npm run build"?

cryptomilk:
I'm a C developer, I don't know npm. I just put the command from deployment/README.md into a Makefile so I don't have to find the command in my history or look it up each time.

Hypfer:
It might be better to just use np run build in the makefile and update deployment/README.md to do this as well so that there is only one place where we possibly have to change things

matthiasharrer:
That was also my intension :) Wasn't meant as an offense @cryptomilk. Just for a brief explanation: npm allows to specify simple commands to be run as an alias which pefectly fits this usecase but it is by far not as powerful as make is. 

As we have to run npm install anyways, I think it males sense to stay inside the npm toolset if possible.

PR #163, opened 2019-04-08
Added final modifications for communication with FloleVac

Flole998:


lib/webserver/WebServer.js
```diff
@@ -616,6 +616,24 @@ const WebServer = function (options) {
...
+        res.status(200).send(self.lastMap.rawData)
+    });
+
+    this.app.get("/api/flolevac/setup", function (req, res) {
+        self.vacuum.getCurrentStatus(function (err, data) {
+            if (err) {
+                res.status(500).send(err.toString());
+            } else {
+                res.json({
+                    token: self.vacuum.token.toString("hex"),
```
onovy: is it good idea to make "token" available to anybody with local network access?

Flole998: You should ask that the person who wrote the function right under this one ;)

Hypfer:
Honestly, I don't quite see the point in providing an interface for an app which does exactly the same that valetudo does while also being closed-source and heavily obfuscated.
It doesn't feel like it belongs in this full-FOSS project.

For a "native-like" app feeling, I suggest that we simply use service workers so that the user can add the Page to his homescreen and all assets are cached on the phone.

fvollmer:
Just an interesting observation: Flole998 created a rebranded fork https://github.com/Flole998/FloleVacWeb/commit/5cae73b8223040f96295130319c612705ce2be0f. The licensing seems like a one way road. I'm not sure why this was necessary, but I'm curious to see if there will be a major feature discrepancy in the future.

cryptomilk:
I think Hypfer's initial response was not very professional. There are already endpoints providing the data he wants to access, they just might not be in the format so they could be directly consumed by his app. He could also just tell the robot to upload the map directly to his app. Parsing the device.conf might be interesting.

He is allowed to fork the project if we wants, the license allows this.

If it is a good idea, I doubt it. Normally it never is ...

Flole998:
@fvollmer Didn't see this until now but no, there won't be any new features introduced, at least it's not planned now. The license change is to prevent PRs or anything against it as (as you might have noticed when you checked my commit history) it's maintained automatically and pulls all changes from Valetudo and appends the necessary modification. So basically the license change makes sure it stays this way and forks/changes are based on Valetudo, especially since I want to prevent development work to be split up. This is not to "destroy" Valetudo by taking the users to an way more advanced project but just for some special use cases that Valetudo doesn't want to be used for.

JokerGermany:
> Just an interesting observation: Flole998 created a rebranded fork [Flole998@5cae73b](https://github.com/Flole998/FloleVacWeb/commit/5cae73b8223040f96295130319c612705ce2be0f). The licensing seems like a one way road. I'm not sure why this was necessary, but I'm curious to see if there will be a major feature discrepancy in the future.

Thats the Reason, why i would use GPL...

PR #170, opened 2019-04-10
Added more Information for MQTT settings in the config.json

thundergreen:


Hypfer:
I'm unsure whether or not the Readme is the right place for this information since it will at some point become very cluttered. I'd say that the wiki is a better place for this.

thundergreen:
The let's ut this in the wiki

PR #175, opened 2019-04-12
Xiaomi coordinates calculations

pszafer:
I took math from Vacuum.prototype.startCleaningZone and copied it to /zone javascript so users can see both coordinates and copy it to different automation systems.


Hypfer:
To quote my response in the telegram group:

Sören, [12.04.19 11:17]
soon™ there shall be support for zoned cleanup using the mqtt vacuum binding

pszafer:
Ok, just wanted to help little bit

Hypfer:
The only thing blocking zoned cleanup via mqtt is the home assistant component not supporting params for send_command so if you know Python it would be awesome if you could fix that :)

PR #178, opened 2019-04-18
Support for mqtt state component in future HA

pszafer:
Added state_topic, support for state string and removed those case.

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -4,12 +4,12 @@ const Tools = require("./Tools");
 const Vacuum = require("./miio/Vacuum");
 
 const COMMANDS = {
-    turn_on: "turn_on",
+    start: "start",
     return_to_base: "return_to_base",
     stop: "stop",
     clean_spot: "clean_spot",
     locate: "locate",
-    start_pause: "start_pause",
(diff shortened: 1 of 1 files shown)


pszafer:
It will be valid from HA 0.93.

----
Note for myself.
I need rename support `turn_on` to `start`

bruvv:
how is this going?

oechslein:
The fan speed 105 / mop is not handled in this code.

pszafer:
Was it supported in earlier versions?

oechslein:
> Was it supported in earlier versions?

You are now able to set "mop"/105 after https://github.com/Hypfer/Valetudo/pull/213 was merged (some days ago). So I assume "FAN_SPEED_TO_STRING[statusData.fan_power]" would then fail, or?
Before your change "mop" wasn't explicitly supported in mqtt. The default branch would have caught it and it would send a "105".

pszafer:
It's different story...
This PR https://github.com/Hypfer/Valetudo/pull/213 added mop only for webui, not mqtt vacuum, that is the reasony why I didn't see it before.
I will create another PR for that.

PR #188, opened 2019-04-26
feat: add MQTT over TLS (mqtts) support

### LINKED ISSUE #126: MQTT over tls?
I would like to use MQTT over tls looking at the [Configuration.js](https://github.com/Hypfer/Valetudo/blob/master/lib/Configuration.js#L13) provided no hints as to if this is possible

timbru31:
Added an option to specify the certificate authority (CA) file path to enable mqtts support. One has to copy e.g. the `mosquitto.crt` file from the MQTT broker to the robo. This is working because the mqtt npm package uses Node.js TLS lib which can use a custom CA list: https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options

Example would be:

`    "caPath": "/mnt/data/valetudo/mosquitto.crt"`

This fixes #126. I've successfully used this setup for some weeks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The Valetudo binary however does not so if you are upgrading your firmware, you 
 * Go-To
 * Zoned Cleanup
 * Configure Timers
-* MQTT
+* MQTT (including TLS support)
 * MQTT HomeAssistant Autodiscovery
 * Start/Stop/Pause Robot
 * Find Robot/Send robot to charging dock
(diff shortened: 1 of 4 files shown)


Hypfer:
It boggles my mind that 

```
const x = "";

if(x) {
  //nothing
} else {
  //this runs.
}
```

Oh JS.

timbru31:
Welcome to JS - https://gist.github.com/MichalZalecki/c964192f830360ce6361 ;)

I found an empty string somehow prettier over null or undefined, if wanted I can change it.

PR #189, opened 2019-04-26
build(TravisCI): execute npm run build command

timbru31:
instead of having another explicit build command


Hypfer:
It boggles my mind that 

```
const x = "";

if(x) {
  //nothing
} else {
  //this runs.
}
```

Oh JS.

Issue #195, opened 2019-04-27
Consumable "Sensor" - Whats meant by that


balu-:
Thanks for this great software,
 I really like it, only one thing took me by surprise:
At the Consumalbes View there is a Consumable called Sensor, whats meant by that?
The other consumables i get, but what is this one?

jowlo:
The sensors are actually not "consumed", but should be cleaned on a regular basis. The same option is exposed in the original Xiaomi App, including instructions. I was asking myself the same question,  and came accross [this thread (in german)](https://www.roboter-forum.com/index.php?thread/20529-meldung-you-need-to-clean-the-sensor-now-was-ist-zu-tun/).

The thread links to two images from the official app, which are pretty self-explanatory. The idea is to just reset the counter after cleaning.
![](https://uploads.tapatalk-cdn.com/20170513/b5e50442b942668427c00a92328337b8.jpg)
![](https://uploads.tapatalk-cdn.com/20170513/b5e50442b942668427c00a92328337b8.jpg)

jowlo:
Would someone (@Hypfer maybe) close this? The issue section is quite messy already...

Hypfer:
@jowlo Mind adding a short explanation to the wiki?

jowlo:
Done.

PR #197, opened 2019-04-28
replaces "Sensor" with "Sensor cleaning"

balu-:
in Replaceable Section of the Settings
to prevent future users from wondering #195  😉


jowlo:
:+1:  If this gets merged, revert `01ca87f` from the wiki (or remove the relevant section).

PR #198, opened 2019-04-29
Limit memory allocation

onovy:
Process valetudo on robot can sometimes consume a lots of memory (memory
leak?). If this happens, kernel sometimes kill player (instead of valetudo
even with high oom score). So let's limit max memory allocation of valetudo
process to 200 MB, which should be enough. Better is to kill valetudo
than player itself.

### DIFF
--- a/deployment/valetudo.conf
+++ b/deployment/valetudo.conf
@@ -8,3 +8,4 @@ oom score 1000
 exec /usr/local/bin/valetudo
 respawn
 respawn limit 10 90
+limit as 209715200 209715200


Hypfer:
This should not be happening. There was a memory leak in node version up to 11.12 iirc. Thats why Valetudo was downgraded to Node 8.

Are you still seeing OOM issues?

Also, 200mb is way too much. Even the current limit of ~80MB is too high.

onovy:
Yes, I am still seeing OOM issues. Even if memory leak is fixed, this is still good idea as "safety check".

Valetudo consumes 145 MB virtual memory when it's idle now in my device. 80 MB limit is just for resources. "limit as" limits virtual memory.

Snímek z 2019-04-29 20-23-54

jowlo:
Well, i don't know too much about how node.js (or how the valetudo packed binary) works internally, but virtual memory has not much to do with actual memory (as in RAM) usage, but includes memory mapped files and all [sorts of other things](https://serverfault.com/questions/138427/what-does-virtual-memory-size-in-top-mean).

You should care about the `RES` column of `top`, which looks quite ok to me...

onovy:
@jowlo you are right, but we don't have option to limit "RES" from top cmd. See man getrlimit. RLIMIT_RSS doesn't work because kernel is >2.4.30 and RLIMIT_DATA doesn't work because kernel is <4.7.

Screenshot of my top is fine, it's not from memory exhaustion situation. Sorry I forgot to take screenshot :). But %MEM (which is % of RES) was 45 %.

Hypfer:
But why would one want to limit virtual memory? 😅

oechslein:
No sure what happens on my version (directly build from github "Zoned_cleanup + Go_to via MQTT"/3b1a9449af6c9fbcf684c63f144e6b137c27167b):

Hypfer:
@oechslein Build command?

oechslein:
> @oechslein Build command?

```
npm install
npm run build
```

onovy:
@Hypfer because we really need to limit max memory allocation and because only way how to do it is virtual memory limit. Yes I know it's not ideal, but much better than oom killer killing player. In last screenshot from @oechslein you can see VIRT=396 MB. So > 200 MB -> kill and start again.

In normal situation virtual memory will not be more than 200 MB.

Hypfer:
You're completely right. Still, I'm surprised that there apparently is a memory leak present.

Can you open another issue for that @oechslein ?

oechslein:
> @Hypfer because we really need to limit max memory allocation and because only way how to do it is virtual memory limit. Yes I know it's not ideal, but much better than oom killer killing player. In last screenshot from @oechslein you can see VIRT=396 MB. So > 200 MB -> kill and start again.
> 
> In normal situation virtual memory will not be more than 200 MB.

Well as soon as I start a zoned cleaning I'm above 200MB VIRT.

oechslein:
> You're completely right. Still, I'm surprised that there apparently is a memory leak present.
> Can you open another issue for that @oechslein ?

Sure thing, but I do not think that it is a memory leak since I do get it right after the start of a zoned cleaning.

onovy:
@oechslein can't reproduce :). Started zoned cleaning and virtual memory is far below 200 MB limit
Snímek z 2019-04-30 12-56-37

Hypfer:
@onovy is it going out of memory?

If not and the map is working, I'd assume that there is something broken in the path handling

onovy:
It's going out of memory after long vacuuming. I just tried reproduce @oechslein problem and failed.

PR #203, opened 2019-04-30
Replaced console prints to simple logger to start improving overall logging

oechslein:
I'm not familiar with node.js so not sure if this is fine. At least it works!
For now the only visible improvement are the added timestamps.


Hypfer:
Why did you choose simple-node-logger?

oechslein:
> Why did you choose simple-node-logger?

I searched for "log" and found its interface most "simple" and sufficient (log into file, onto console, rolling files). I do not have experience with npm at all.

oechslein:
@Hypfer: Will close it, just reopen if you consider adding it.

PR #205, opened 2019-05-01
Changed the upstart file to work when unprovisioned (AP mode)

knobunc:
This change made my vacuum bring up Valetudo when unprovisioned
(i.e. when it had not joined a host network and was running the AP
roborock-vacuum-s5_... one itself).


Hypfer:
Will this reintroduce firmware rollbacks? We'll see.

PR #212, opened 2019-05-05
Add a basic HTTP status notifier

onfoot:
There are no means of getting Vacuums status without polling it constantly or setting up MQTT, so I implemented a basic status notifier that `POST`s the status over HTTP to a URL defined in the configuration.

Example entry in `/mnt/data/valetudo.config.json`:

```
"http-status-notifications": {
    "url": "http://192.168.0.116:52199/status"
}
```

I admit this pull request is a little selfish, because I need such functionality for [my own thing](https://github.com/onfoot/homebridge-valetudo-xiaomi-vacuum) - I hate to bother the poor overloaded vacuum with constant status polls, so a push would be nice instead. ;)

It could also be used to send push notifications through e.g. Pushover or execute webhooks in IFTTT. I know Valetudo is all about freeing the Vacuum from the "cloud", but we can make it work in the "clouds" we kind of control. ;)

If you like it, and have any ideas on how to make it more complete, or perhaps more stable and crash resistant, let me know!

### DIFF
--- a/lib/HttpNotifier.js
+++ b/lib/HttpNotifier.js
@@ -34,14 +34,21 @@ HttpNotifier.prototype.postStatus = function(statusData) {
         headers: {'Content-Type': 'application/json'}
     };
 
-    const req = http.request(options, (res) => {});
-    
+    let req = null;
     try {
+        req = http.request(options);
+        req.on('error', () => {});
+
         const stringified = JSON.stringify(statusData);
-        req.write(stringified);
-    } catch (e) {
-    } finally {
+        if (stringified) {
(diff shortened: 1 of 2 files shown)


Hypfer:
While I really appreciate contributions, I'm sorry to say that this won't be merged.
MQTT already does what you want to do and is a lot more flexible in doing so.

If you require the use of webhooks, you could build a small application using something like [this](https://github.com/mcollina/mosca) which takes the mqtt messages and calls your webservice.
Just package it like valetudo and also run it on the robot. :)

PR #213, opened 2019-05-07
align fan speeds with the xiaomi cleaning modes

leonardpitzu:
change the fan speed to the original cleaning modes
add the mop mode (still needs some checking)


onovy:
maybe is better to explain in commit message what you did. "Update index.html" sounds really useless. "Align fan speeds with the Xiaomi cleaning modes" sounds much better :)

PR #219, opened 2019-05-08
Added "mop mode" to fan speed dropdown. Sent value 105 to fanspeed API

rschueler:
Need to add handling for gen1 to disable or don´t show this option, because it is only available on gen2 robots.


rschueler:
Has been added already. So you can ignore this.

Issue #165, opened 2019-04-09
Voicepacks


Hypfer:
Since there seems to be some demand for custom voicepacks, I'd like to write down my thoughts on how Valetudo should approach this topic.

It might make sense to split this into multiple phases:

**Phase 1**
~~A simple file upload dialog where the user can pick a local language pack file and upload it through valetudo to the robot.~~

**Phase 2**
In the long run, I'd love to see a seperated github repo (maybe *valetudo-voice*) which contains curated and free voicepacks. Valetudo should fetch the latest list of all available voicepacks in said repo and render a nice view where the user can simply tap on one to install it as well as play a preview of it.

There should be a template voicepack for *valetudo-voice* so that people who want to contribute a new pack know which files they need and what they do. Then do some Travis CI stuff on Pull Requests to build the voicepack archives in the robot format.

thundergreen:
I totally agree. We might add other languages than English and German. Like French Italien etc.

Hypfer:
Since the language of the voicepack is controlled by `/mnt/default/roborock.conf` I wonder what happens if you change `language=prc` to `language=someRandomString`.

If that just loads the files from someRandomString, we might have a way to manage multiple packs for a single language or packs for no real languages at all.

TheRealMoeder:
Hi! I already realized an automated voicepack generator for a different project utilizing pythonanywhere, github and travis. As it was only little work, I quickly ported this to be used with valetudo. The voices used are mostly high quality versions from MacOS X, so the result is pretty good with many languages available. You can try the service [here](http://moeder.pythonanywhere.com/rockrobo), the resulting pkgs are pushed back to my [github.io page](http://therealmoeder.github.io/rockrobo.html). It usually takes only 2-3 minutes for the voicepack to get finished.

TheRealMoeder:
@Hypfer thanks for the thumbs up ;) Do you think the voice pack generator is something which should be more refined for valetudo? I could try to spare some time. The method of python triggering Travis CI is pretty usable, but I still dislike the pushback to GitHub.io, as it is more of a hack. I guess I could also store the files on pythonanywhere and give a hashed link for download, but filestorage and bandwidth are very limited on the free plan. I haven't found any other suitable free service. Any recommendations?

OzGav:
Hi,
[...]

PR #239, opened 2019-05-20
Install voice packs through valetudo

r1sim:
This pull request adds an option in the sound settings to upload and install voice packs.
These changes implement "Phase 1" of #165.

**I added ~~3~~ 2 new express endpoints:**
- _POST - /api/install_voice_pack_ - To upload the pkg file
- _GET - /api/install_voice_pack_status_ - To get the current voice pack installation status
~~- _GET - /api/download_voice_pack_ - miio wants some URL to download the language pack from. This endpoint will basically just provide the file that was uploaded using the POST endpoint. For the sake of security it is not possible to access this endpoint from remote addresses in production mode.~~

**I also added 2 new functions in Vaccum.js:**
- _installVoicePack(url, md5, callback)_ - To start the voice pack installation
- _getVoicePackInstallationStatus(callback)_ - To get the status of the voice pack installation

Tested on Gen 2.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -2517,13 +2517,16 @@
                         }, function (err, res){
                             if(err){
                                 ons.notification.toast(err, { buttonLabel: 'Dismiss', timeout: 1500 });
+                                voicePackUploadButton.innerText = uploadText;
+                                voicePackUploadButton.disabled = false;
+                                loadingBarSettingsSoundVolume.value = 0;
                             }else{
                                 voicePackUploadButton.innerText = 'Installing voice pack...';
                                 getVoicePackInstallStatus(function(err, data){
                                     if(!err){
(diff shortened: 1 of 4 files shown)


Hypfer:
Nice work.
We do need to delete the uploaded voicepack as soon as it is installed, though.

Otherwise it will fill up the RAM, crash the robot, cause firmware rollbacks etc.

r1sim:
Thats a good point. I just did some additional changes:

- The temporary upload files are now stored at /mnt/data/valetudo/uploads because this partition has way more free space than /tmp.
- Uploads will be deleted one minute after installing the voice pack. If that doesn't work for whatever reason, the uploads folder will also be cleared on every start of valetudo and before every voice pack upload.
- I discovered that it is not necessary to provide the file internally via http as the miio function to install voice packs also accepts "file://" urls. Therefore i removed the /api/download_voice_pack endpoint.

PR #241, opened 2019-05-30
Fixed vacuum states for mqtt client.

jurenovic:


dlashua:
I've not tested this PR. However, based on the code in Home Assistant, I don't believe "zoned cleanup" will be recognized as "cleaning" without a template in the MQTT Sensor config. PR #242 sets the state to "cleaning" during a zoned cleanup as well, which will be recognized properly. Without either a template or using states it's aware of, HASS will keep the vacuum in the previous state.

Hypfer:
What does this PR do?

jurenovic:
@dlashua I agree it doesn't match the MQTT vacuum states directly in home-assistant. I am using appdaemon to schedule next zone cleanup after the first one has ended. Before that i was polling the status api. Now with MQTT i am just waiting for that state to be pushed back.
@Hypfer What I wanted to achieve with this PR is to have the same state pushed trough MQTT as the one in Valetudo status api call.
Feel free to close it if not needed.

Hypfer:
It seems like the current mapping is the same as the one found in the xiaomi miio vacuum component?
https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/xiaomi_miio/vacuum.py

So.. will this break things? Is there some list of "correct" possible status?
If not, having more information would definitely be a good enhancement.

Hypfer:
Okay, so apparently we have to be compatible to the possible states defined here:
https://github.com/home-assistant/home-assistant/blob/master/homeassistant/components/vacuum/__init__.py#L58

This means that this PR cannot be merged as it will break HA compatibility.

https://github.com/Hypfer/Valetudo/commit/6b304f090b555d6a2cb5445e1a275654bf51e406
I did however add another property to the attribute topic which contains the infos you need.
image

jurenovic:
@Hypfer I am fine with having another field as `valetudo_state` inside mqtt, but then let's map it to same states that the `/api/current_status` returns.  [Vacuum states](https://github.com/Hypfer/Valetudo/blob/master/lib/miio/Vacuum.js#L629-L647)

Hypfer:
Thats exactly what I did

jurenovic:
Ahh, now i see it, sorry for not taking a second look before replying. 👍

Issue #255, opened 2019-06-06
Fan Power matches only exact values, other values show undefined in GUI


sklerotraficon:
Fan Power values rely on fixed value steps to match the labels shown in the GUI:
`var fanspeedPresets = {
                    38: "Quiet",
                    60: "Balanced",
                    75: "Turbo",
                    100: "Max",
                    105: "Mop"
                };
`

Since you may use other tools, like miio or [RoboRock Control Center](https://github.com/LazyT/rrcc), that allow other values than the above presets, this will result in "undefined" being shown in the GUI.

This should be corrected, that the fanpower is matched in ranges of the predefined steps.

A nice to have would be an additional "Whisper" preset, which turns the fan (almost) off, this is actually useful for moving the robot around with "GoTo".

PR #257, opened 2019-06-06
Issue 255 Fan Power matches only exact values, other values show undefined in GUI

sklerotraficon:
This PR resolves the issue described in #255

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -358,14 +358,6 @@
             <script>
                 var currentRefreshTimer;
 
-                var fanspeeds = {
-                    38: "Quiet",
-                    60: "Balanced",
-                    75: "Turbo",
-                    100: "Max",
-                    105: "Mop"
-                };
-
                 var startButton = document.getElementById("start-button");
                 var pauseButton = document.getElementById("pause-button");
                 var stopButton = document.getElementById("stop-button");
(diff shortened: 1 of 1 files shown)


Hypfer:
Shouldn't it show the value itself instead of the closest one available?

sklerotraficon:
I'm not really sure, what you mean, the current fan power is afaik never displayed in the GUI, only the preset labels.

Hypfer:
https://github.com/Hypfer/Valetudo/pull/257/files#diff-7f958a715ef0d644371ae8b1b5f18250R487

This maps a fanspeed which isn't the preset to the preset?

sklerotraficon:
Is it the method's naming that your comment is about ? Thnx for being a little more specific!

Hypfer:
Nah.

We should show the real fanspeed value as a number if it isn't a preset instead of mapping it to the nearest valid preset which I assume the code I've referenced is doing.

Haven't tested it, though. That's why it might've been confusing

sklerotraficon:
Yes, that was my intention:

If the fanpower value is lower or equal the threshold in the preset, use the preset threshold's label:

Example: 
Fanpower value greater than 38 and lower or equal 60 is showing "Balanced".

I actually like the presets, but there are options completely without them:

1. Fanpower Value only, eg. 50
2. Fanpower in Percent, eg. 50%
3. Volume-like control

Otherwise, with presets, it could be confusing to have both, numeric and presets on the button... don't you think?

Hypfer:
The problem with this approach is that the UI just lies, because the fan speed is in fact not Balanced but rather something inbetween. This is imo worse than just showing undefined.

However I understand your point. I'd suggest showing something along the lines of `Custom: 50%`

sklerotraficon:
Updated the PR:

If the value is not exactly matching the exact value from the preset, then the following is displayed:

Custom [fanpower]%

PR #261, opened 2019-06-10
Implement https support out of Valetudo

Poeschl:
I implemented a new config section with ssl settings:
```
 "ssl": {
    "enabled": false,
    "key": "/mnt/data/valetudo/roborock.key",
    "cert": "/mnt/data/valetudo/roborock.crt"
  }
```
and also the function of it. If the ssl toggle is switched to `true`. Valetudo will serve its website via `https://` on port `443`.

The frontend behaves very nice and sends all requests now to the secure frontend, but for some reason the map isn't getting any updates anymore. The "No mapdata" map stays there, even if the robot is moving.
Since I'm out of ideas, I hope someone can help me with the map refresh.

### DIFF
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -28,8 +28,13 @@ const Configuration = function() {
             username: "valetudo",
             password: "valetudo"
         },
-	"allowSSHKeyUpload": true,
-        "map_upload_host": "http://127.0.0.1"
+        "allowSSHKeyUpload": true,
+        "map_upload_host": "http://127.0.0.1",
+        "ssl": {
+            enabled: false,
+            key: "/mnt/data/valetudo/key.key",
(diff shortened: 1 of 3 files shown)


rand256:
You probably should change `map_upload_host` value from http:// to https:// in configuration, or make webserver listen to both http and https.

Poeschl:
I gave the `map_upload_host` a quick try. Unfortunately it didn't seem to wok. 
I might give the communication between the dustcloud and the frontend a look the next days

rand256:
Well, it may have some checks for certs being valid, or just can't work via https at all. So for those local connections it might be better to use just plain http requests to valetudo webserver.

Also, after the issue with no map data will be solved, you also should check websockets to use wss:// instead of ws://.

howsTricks:
Wondering what the status is of this PR?

PR #272, opened 2019-06-19
simplify map saving

greenimpala:
Screenshot 2019-06-19 at 17 56 15

Currently in `master` map saving is not working correctly, `set_lab_status` is not enough to enable map persistence and the UI could be confusing to users without knowledge of `mirobo` commands.

Following on from https://github.com/Hypfer/Valetudo/pull/247 — this PR simplifies the UI with two buttons:

* `Save Current Map` ->  Calls raw command `set_lab_status 1` + `save_map`.
* `Reset Current Map` ->  Calls raw command `reset_map`.

Note that calling `reset_map` is enough to disable map saving.

Tested extensively on an S55.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -1682,7 +1682,7 @@
             </ons-card>
             <ons-card onclick="fn.pushPage({'id': 'settings-persistent-data.html', 'title': 'Persistent Maps'})">
                 <div class="title"><ons-icon icon="fa-map"></ons-icon> Persistent Data</div>
-                <div class="content">Configure the lab mode for enabling virtual walls etc.</div>
+                <div class="content">Saves the map, zones and virtual walls.</div>
             </ons-card>
             <ons-card onclick="fn.pushPage({'id': 'settings-consumables.html', 'title': 'Consumables'})">
                 <div class="title"><ons-icon icon="fa-wrench"></ons-icon> Consumables</div>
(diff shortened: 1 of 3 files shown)


hagenuck1:
If this is only a new button with the same functionality as „enable“ before I guess this is more irritating. 

For me „save current map“ look more like to backup the current map to reload it back later.

greenimpala:
The map saving does not work for me unless `save_map` is called. Checking the `lab_status` and showing a toggle in the UI is not (to me) a good way of showing whether map saving is working as you still need to issue further commands to enable persistence.

Hypfer:
So now there is no way to disable lab mode?

hagenuck1:
@greenimapala are you sure that „user_map0“ is there if persistent Maps are active? 

I‘m having them active and reload my maps from my „last_map“ file and didn’t save the user_map any time.

greenimpala:
> So now there is no way to disable lab mode?

Correct. Maybe best of both would be a lab mode toggle and (when lab mode is on) have the two buttons save and reset.

cryptomilk:
I'm sorry, but this seems to be wrong.

`save_map` is used to save persistent map data, barriers and no-go zones! Just calling `save_map' with any parameters means it removes all barriers and no-go zones. Normally you need to parse the current map data and then e.g. remove a barrier and save it again. Passing all the data you don't want to remove.

greenimpala:
> I'm sorry, but this seems to be wrong.
> 
> `save_map` is used to save persistent map data, barriers and no-go zones! Just calling `save_map' with any parameters means it removes all barriers and no-go zones. Normally you need to parse the current map data and then e.g. remove a barrier and save it again. Passing all the data you don't want to remove.

That makes sense that barriers etc are updated with `save_map` but just enabling lab mode was not enough to get the actual map itself to persist. I am using this fork successfully but I will close the PR as it is causing some confusion and it could be a weird issue on my end.

matthiasharrer:
What about also saving the map when enabling labmode, so that it does not rotate? I think this should not cause any problems as there should not be any persistent data when labmode is disabled, right?

In addition we will have a persistent data editor with the save_map functionality in the future of course.

greenimpala:
> What about also saving the map when enabling labmode, so that it does not rotate? I think this should not cause any problems as there should not be any persistent data when labmode is disabled, right?

Yup that's what this PR did.

matthiasharrer:
Well not really ..
I think the save map button will be clickable also when lab mode is already turned on and therefore potentially delete all the virtual walls and stuff.
However only saving the map when we know that lab mode was turned off before (and there are no virtual walls) should still be possible and probably also a good idea?

Editing and saving the map later on is already implemented here https://github.com/rand256/valetudo but I'm not sure if @rand256 actually wants to merge this back..

PR #277, opened 2019-06-26
Made the docker image much smaller

bruvv:
Also made it much simpelere and do not clutter it.

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -3,7 +3,7 @@ FROM node:10.16.0-alpine
 WORKDIR /deploy
 
 # make sure you add the volume to the docker container /build to your host map
-# SSH to your vacuum, stop the valetudo service (service valetudo stop), replace /usr/local/bin/valetudo with your new binary and restart valetudo (service valetudo start).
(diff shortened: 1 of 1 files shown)


deployment/Dockerfile
```diff
@@ -1,17 +1,23 @@
...
-; n latest \
-; git clone http://github.com/hypfer/Valetudo \
-; cd Valetudo \
-; /usr/local/bin/npm install \
-; /usr/local/bin/npm run build \
+; npm install -g pkg \
+; mkdir "/deploy" \
+; echo "#!/bin/sh" >> /deploy/build.sh
+
+RUN echo "git clone https://github.com/Hypfer/Valetudo.git tmp && mv tmp/.git /deploy && rm -rf tmp && git reset --hard" >> /deploy/build.sh
```
onovy: running less RUN(s) is much faster in docker build.

Why not just COPY build.sh file from hostmachine?

bruvv: True about the RUN(s). Will change it. And I did not want to clutter the github anymore with useless files. We only need the build.sh in the docker so we do not need to add it to the git in my eyes. :) 

God points tho! Thanks for the feedback :)

PR #281, opened 2019-07-02
make virtual walls and forbidden zones user-configurable

rand256:
a bit stripped version of the same functionality from Valetudo RE fork separated from most customizations and tweaks of the mod

this one not tested on real hardware but hopefully should work

### DIFF
--- a/client/forbidden-markers-configuration-map.html
+++ b/client/forbidden-markers-configuration-map.html
@@ -1,0 +1,106 @@
+<ons-page id="forbidden-markers-configuration-map-page">
+    <div id="forbidden-markers-configuration-container">
+        <div class="forbidden-markers-configuration-header">
+            <ons-back-button>Forbidden Markers</ons-back-button>
+            <span id="forbidden-markers-configuration-map-page-h1"></span>
+        </div>
+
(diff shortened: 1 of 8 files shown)


bruvv:
What is the Valetudo RE fork you are talking about?

rand256:
This one: https://github.com/rand256/valetudo
Was discussed on telegram channel.

leonardpitzu:
hi. i adapted your PR to my code base. it works as expected but i have one question - is there a way to edit the wall/forbidden area?

rand256:
It works just the same way as editing zones.

Hypfer:
While this is already a huge improvement, I'd still prefer valetudo to save the no-go areas independently instead of just calling save_map, because there might be breakage.

Calling save_map should definitely trigger a map poll on the cloud interface, since closing the editor and opening it again is sometimes displaying stale data in my tests. I'll add that.

Anyways thank you for this PR. If you feel like it, I'd be glad to see the other valetudo-re features merged into master as well.

Hypfer:
Sadly, I'm unable to delete no-go zones as well as virtual walls :/

Issue #282, opened 2019-07-04
Question: Map parsing in MapDTO? / Unterstanding


balu-:
Not a issue just a question.

I was browsing though the files and trying to understand the working of Valetudo.
So i came to the map & map-parsing parts (so about the  Valetudo/lib/RRMapParser.js and  Valetudo/lib/dtos/MapDTO.js files). 
The question that came up was, shouldn't the map parsing logic reside in the MapDTO? 
As i understand it:
The robo software transports map data via http call to the valetudo software, 
which parses this data an holds it in the MapDTO. 
-> In my mind the Data Transport object should not only hold the data but also the capability to "pack" & "unpack" it  (as packing is not needed only "unpack" as in parse)

Or is there anything wrong with my logic/ am i missing something?

Hypfer:
Sounds reasonable 🤔
In its current state, the DTO actually doesn't make much sense at all.

I'd say go for it

PR #292, opened 2019-07-17
moves MapParsing to MapDTO

balu-:
see #282 
didn't test the code 😞


Hypfer:
Since you didn't attach it anywhere, this won't work because PARSE is not exported. It's just a const which never gets accessed.

balu-:
I thought it would be part of the dto (cause it’s inside the function and thereby be implicitly exported. But I’ll read up on this and test the code

Hypfer:
This will be superseded by the rewrite for supporting other robots

Anyways thank you for your efforts

PR #316, opened 2019-08-31
Fix #289 Instantly update no-go zones and virtual walls

### LINKED ISSUE #289: Change of No-Go Zones/Virtual walls isn't instantly visible
For some reason, propagation of this takes some time.
If you close the editor and reopen it right again, you'll see stale data.

r1sim:
This PR fixes #289.

The issue was caused because the no-go zones and virtual walls are stored in xiaomis map data. Valetudo caches the map data in the `self.map.parsedData` variable in WebServer.js.
This variable only gets updated when the robot uploads a new map to the dummy cloud.

Valetudo calls the `/api/map/latest` endpoint before opening the editor window to get the latest no-go zones and virtual walls. The information returned by this endpoint might not be up-to-date when the no-go zones where changed after the last map upload.

To fix this we just needed to update the `self.map.parsedData` variable in the `/api/persistent_data` endpoint with the data send by the frontend (used to update no-go zones and virtual walls). This will make sure that the changes are visible even before the robot has uploaded a new map.

### DIFF
--- a/lib/webserver/WebServer.js
+++ b/lib/webserver/WebServer.js
@@ -550,6 +550,8 @@ const WebServer = function (options) {
                     if(err) {
                         res.status(500).json(err);
                     } else {
+                        self.map.parsedData.virtual_walls = req.body.virtual_walls;
+                        self.map.parsedData.no_go_areas = req.body.no_go_areas;    
                         res.status(201).json({message: "ok"});
                     }
                 });


rand256:
This way you won't know whether your changes were successfully applied or not.

r1sim:
I know it would be better to somehow read this information back from the robot. Unfortunately I couldn't find any documentation on how to do this, so I just assumed that the no-go zones were set if the miio command to do this returned without any errors. 

Maybe we could somehow trigger a map upload after updating the changed no-go zones, but I guess that this would still take some time to populate the changed values. Any thoughts on that?

rand256:
Surely you can't find any documentation since it doesn't exist.

Map upload is [already triggered](https://github.com/Hypfer/Valetudo/blob/master/lib/miio/Vacuum.js#L655).

Hypfer:
What's the status on this?

r1sim:
I don't have any updates on this. I guess I'll close this PR and see if I can come up with a better solution.

Issue #69, opened 2018-12-29
Easy way to store and restore Map for Gen1 & Gen2


ovaltineo:
I'm not sure if anyone has done this before (maybe I didn't look hard enough) but it was pretty easy to store and restore the current map on a rooted Xiaomi robot.  To store, I just copied /mnt/data/rockrobo/* (excluding directories) to a backup folder.  To restore, I copied from the backup folder to /mnt/data/rockrobo, then I kill 'player' process which gets restarted by the watchdog within a minute.  Voila, the robot now starts using the restored map.

I think these two functions can easily be incorporated in Valetudo.  It certainly saves me from re-mapping the house whenever the robot decides to forget the map.  My zoned cleanup coordinates are always spot on because of this.

roflcoopter:
> mnt/data/rockrobo

Are you doing this consistently to have separate maps? 
Are you deleting the contents of /mnt/data/rockrobo/ and then swapping in different maps?

ovaltineo:
I am using it to restore the map whenever the robot gets stuck and someone returns it manually to the base.  If there are two saved folders, then map swapping should work.  I don't need to delete the files because they are copied over from the backup.  I use a remote SSH script to do the copying.  This script is triggered via Google Home voice command configured in IFTTT.

roflcoopter:
Interesting! Mind sharing your script?

ovaltineo:
Here's the script run on my raspberry pi for copying the files from backup.  ssh key is used so no there are no password prompts.  I use Blynk to trigger this script.  IFTTT sends message to Blynk.

##restore_rockrobo.sh
MIROBO_IP=192.168.1.15
ssh root@$MIROBO_IP cp /mnt/data/rockrobo/backup/* /mnt/data/rockrobo;
ssh root@$MIROBO_IP sync
ssh root@$MIROBO_IP reboot

TheMicroDevil:
Just the hint i was looking for, thanks!
I've  developed a tool in bash, which can issue single/multiple zone cleaning (running on device or wherever). This could be the missing function to reset the map before start.

To be mad,  theoretically it could then support multiple floors too, if needed.
Something like:
clean $zone@$floor
- if $floor not $standard_floor
- play sound (hey, you have to do something)
- wait for error 3
- you take the robot to $floor, error 3 => 'Wheel floating', sound stops
- wait for error 0 - no error, the robot is there
[...]

Issue #91, opened 2019-02-11
Multiple Maps


sirtet:
Questions about Idea for a new Feature:
I am trying to decide which robot t buy. 
As i understand, they all only support a single map, so are not comfortably usable for non-connected rooms Like different floors, right? 

Reading here, i wonder if it Would be possible to implement multi rooms by eg:
Let the robot record a map of every area, save it somewhere and add some Interface to swap maps whenever you carry it to the New area? Web UI or maybe even via a Hardware mod, Like a nfc sensor/bt beacon? 

Or, could you simply stitch the different maps together? I guess the robot Would happily just clean the area it can currently reach? 

Not being able to reach the charger base will always be an issue i guess. 
Do they send an SOS if they are unable to dock? 

How unrealistic am i?

cryptomilk:
See #86 and contribute :-)

Hypfer:
Since even Gen2 robots are currently priced around 320€, hacking the firmware so that it supports map swapping isn't reasonable at all. It's significantly cheaper to just buy as many robots as there are floors instead of spending weeks on this feature.

Even if you'd manage to get it to work it would still introduce lots and lots of new bugs since the roborock software is just not made for this use-case.
Therefore this is not and will not be supported by Valetudo

sirtet:
> See #86 and contribute :-)

will try to if i get a robot, but i'll need tasks other than coding...

> ... It's significantly cheaper to just buy as many robots as there are floors instead of spending weeks on this feature.

Good point, but

- my old house has 3 floors and on 2nd floor 4 non-connected rooms, aka 6 robots...
- one might just not want a robot and dock in every room, for space, esthetics or whatever.
- one might want to save energy and physical resources, even if they are cheap to buy.

> Even if you'd manage to get it to work it would still introduce lots and lots of new bugs since the roborock software is just not made for this use-case.
> Therefore this is not and will not be supported by Valetudo

Fair enough. I just saw #69 also mentions the use of multiple floors, so i thought it might be common practice.

I don't know anything about vacuums yet, just trying to decide what to buy. 
But as i like hackability, i guess i'll go for a roborock s5 and see what happens.

treksler:
This was the only feature holding me back from Getting a Roborock over the Neato, which supports saving maps for up to 3 floors. Thank you for pointing me to the instructions in #69 
They look trivial enough.

[...]

PR #317, opened 2019-09-06
adding multi floor support

ppisljar:
resolves https://github.com/Hypfer/Valetudo/issues/69

adds two buttons to the home panel:
- clicking save map asks for a name, then it copies the map to a backup folder with provided name
- clicking load map lists all the saved maps you created and allows you to restore one

it works the way @hagenuck1 described in the original issue. as follow up to this i hope to get it working without the reboot.

it should also work over MQTT custom commands "load_map" and "save_map" but i haven't tested those.

image

### DIFF
--- a/lib/MapManager.js
+++ b/lib/MapManager.js
@@ -18,7 +18,8 @@ function storeMap(name, cb) {
     else {
         // we don't have charger info
     }
-    fs.copyFileSync("/mnt/data/rockrobo/PersistData_1.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_1.data");
+    if (fs.existsSync("/mnt/data/rockrobo/PersistData_1.data")) fs.copyFileSync("/mnt/data/rockrobo/PersistData_1.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_1.data");
+    if (fs.existsSync("/mnt/data/rockrobo/PersistData_2.data")) fs.copyFileSync("/mnt/data/rockrobo/PersistData_1.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_2.data");
     cb(null);
 }
 
@@ -27,12 +28,14 @@ function storeMap(name, cb) {
  * @param name
  */
 function loadMap(name, cb) {
+    if (fs.existsSync("/mnt/data/rockrobo/PersistData_1.data")) fs.unlinkSync("/mnt/data/rockrobo/PersistData_1.data");
     if (fs.existsSync("/mnt/data/rockrobo/PersistData_2.data")) fs.unlinkSync("/mnt/data/rockrobo/PersistData_2.data");
     if (fs.existsSync("/mnt/data/rockrobo/StartPos.data")) fs.unlinkSync("/mnt/data/rockrobo/StartPos.data");
     if (fs.existsSync("/mnt/data/rockrobo/user_map0")) fs.unlinkSync("/mnt/data/rockrobo/user_map0");
     fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/last_map", "/mnt/data/rockrobo/last_map");
     if (fs.existsSync("/mnt/data/valetudo/maps/" + name + "/ChargerPos.data")) fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/ChargerPos.data", "/mnt/data/rockrobo/ChargerPos.data");
-    fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/PersistData_1.data", "/mnt/data/rockrobo/PersistData_1.data");
+    if (fs.existsSync("/mnt/data/valetudo/maps/" + name + "/PersistData_1.data")) fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/PersistData_1.data", "/mnt/data/rockrobo/PersistData_1.data");
+    if (fs.existsSync("/mnt/data/valetudo/maps/" + name + "/PersistData_2.data")) fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/PersistData_2.data", "/mnt/data/rockrobo/PersistData_2.data");
     fs.readFile("/mnt/data/rockrobo/RoboController.cfg", "utf8", function (err, data) {
         if (err) {
             console.log(err);
@@ -44,7 +47,7 @@ function loadMap(name, cb) {
             if (err) {
                 return cb(err);
             }
-            exec("reboot");
+            exec("service rrwatchdoge restart");
             cb(null, "ok");
         });
     });
--- a/lib/MapManager.js
+++ b/lib/MapManager.js
@@ -14,7 +14,10 @@ function storeMap(name, cb) {
         fs.mkdirSync("/mnt/data/valetudo/maps/" + name);
     }
     fs.copyFileSync("/mnt/data/rockrobo/last_map", "/mnt/data/valetudo/maps/" + name + "/last_map");
-    fs.copyFileSync("/mnt/data/rockrobo/ChargerPos.data", "/mnt/data/valetudo/maps/" + name + "/ChargerPos.data");
+    if (fs.existsSync("/mnt/data/rockrobo/ChargerPos.data")) fs.copyFileSync("/mnt/data/rockrobo/ChargerPos.data", "/mnt/data/valetudo/maps/" + name + "/ChargerPos.data");
+    else {
+        // we don't have charger info
+    }
     fs.copyFileSync("/mnt/data/rockrobo/PersistData_1.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_1.data");
     cb(null);
 }
@@ -28,7 +31,7 @@ function loadMap(name, cb) {
     if (fs.existsSync("/mnt/data/rockrobo/StartPos.data")) fs.unlinkSync("/mnt/data/rockrobo/StartPos.data");
     if (fs.existsSync("/mnt/data/rockrobo/user_map0")) fs.unlinkSync("/mnt/data/rockrobo/user_map0");
     fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/last_map", "/mnt/data/rockrobo/last_map");
-    fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/ChargerPos.data", "/mnt/data/rockrobo/ChargerPos.data");
+    if (fs.existsSync("/mnt/data/valetudo/maps/" + name + "/ChargerPos.data")) fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/ChargerPos.data", "/mnt/data/rockrobo/ChargerPos.data");
     fs.copyFileSync("/mnt/data/valetudo/maps/" + name + "/PersistData_1.data", "/mnt/data/rockrobo/PersistData_1.data");
     fs.readFile("/mnt/data/rockrobo/RoboController.cfg", "utf8", function (err, data) {
         if (err) {
--- a/lib/MapManager.js
+++ b/lib/MapManager.js
@@ -13,6 +13,9 @@ function storeMap(name, cb) {
     if (!fs.existsSync("/mnt/data/valetudo/maps/" + name)) {
         fs.mkdirSync("/mnt/data/valetudo/maps/" + name);
     }
+    if (!fs.existsSync("/mnt/data/last_map")) {
+        return cb("map doesn't exist");
+    }
     fs.copyFileSync("/mnt/data/rockrobo/last_map", "/mnt/data/valetudo/maps/" + name + "/last_map");
     if (fs.existsSync("/mnt/data/rockrobo/ChargerPos.data")) fs.copyFileSync("/mnt/data/rockrobo/ChargerPos.data", "/mnt/data/valetudo/maps/" + name + "/ChargerPos.data");
     else {
@@ -28,6 +31,10 @@ function storeMap(name, cb) {
  * @param name
  */
 function loadMap(name, cb) {
+    if (!fs.existsSync("/mnt/data/valetudo/maps/" + name + "/last_map")) {
+        cb("floor '"+cb+"' doesn't exist");
+        return;
+    }
     if (fs.existsSync("/mnt/data/rockrobo/PersistData_1.data")) fs.unlinkSync("/mnt/data/rockrobo/PersistData_1.data");
     if (fs.existsSync("/mnt/data/rockrobo/PersistData_2.data")) fs.unlinkSync("/mnt/data/rockrobo/PersistData_2.data");
     if (fs.existsSync("/mnt/data/rockrobo/StartPos.data")) fs.unlinkSync("/mnt/data/rockrobo/StartPos.data");
--- a/lib/MapManager.js
+++ b/lib/MapManager.js
@@ -13,7 +13,7 @@ function storeMap(name, cb) {
     if (!fs.existsSync("/mnt/data/valetudo/maps/" + name)) {
         fs.mkdirSync("/mnt/data/valetudo/maps/" + name);
     }
-    if (!fs.existsSync("/mnt/data/last_map")) {
+    if (!fs.existsSync("/mnt/data/rockrobo/last_map")) {
         return cb("map doesn't exist");
     }
     fs.copyFileSync("/mnt/data/rockrobo/last_map", "/mnt/data/valetudo/maps/" + name + "/last_map");
--- a/lib/MapManager.js
+++ b/lib/MapManager.js
@@ -1,0 +1,62 @@
+const fs = require("fs");
+const { exec } = require("child_process");
+
+/***
+ * stores current map to a backup folder
+ * @param name
+ */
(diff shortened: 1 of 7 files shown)


hagenuck1:
Hi, see my updates here:
https://github.com/hagenuck1/Valetudo/wiki/Load-saved-maps-using-Home-Assistant

I found out a reboot isn't needed! But you have to "service rrwatchdoge restart" and wait 30 Seconds for the new map to be loaded. If you start cleaning within this time robot will reboot and may reset!

lib/MqttClient.js
```diff
@@ -419,6 +422,40 @@ MqttClient.prototype.handleCustomCommand = function (message) {
                     console.info("Missing spot_id");
                 }
                 break;
+
+            /**
+             * {
+             *   "command": "go_to",
+             *   "spot_id": "Somewhere"
+             * }
```
hagenuck1: Comment should be edited:
/**
             * {
             *   "command": "store_map",
             *   "name": "Somemapname"
             * }
             */

lib/MqttClient.js
```diff
@@ -419,6 +422,40 @@ MqttClient.prototype.handleCustomCommand = function (message) {
...
+                    MapManager.storeMap(msg.name, function(err, data) {
+                        if (err) { console.info(err); }
+
+                    })
+                } else {
+                    console.info("Missing name");
+                }
+                break;
+
+            /**
```
hagenuck1: Comment should be edited:
/**
             * {
             *   "command": "load_map",
             *   "name": "Somemapname"
             * }
             */

lib/MqttClient.js
```diff
@@ -14,7 +15,9 @@ const MQTT_COMMANDS = {
 
 const CUSTOM_COMMANDS = {
     GO_TO: "go_to",
-    ZONED_CLEANUP: "zoned_cleanup"
+    ZONED_CLEANUP: "zoned_cleanup",
+    LOAD_MAP: "load_map",
+    STORE_MAP: "store_map",
```
hagenuck1: comma after "store map", should be needed?

ppisljar:
thanks for the feedback, i'll update based on your guidelines

i have some quick questions, where answers might help me make this work better:
- can we get the current position ? (x/y coordinates where the robot currently is in the map) 
- and can we change/update the current position ? (start point in your map?)
- is there a way to detect that robot is lifted off the ground ?

hagenuck1:
I haben tried that so far, but the robot will detach its position automatically when the map and the rrwatchdoge are restarted and you start cleaning/ goto. 

I don’t know of any way to get known about lifting up, sorry.

ppisljar:
if robot is lifted and you try to perform an action it signals an error ... possible improvement:

- with each floor set the 'stairs' location, this is the location where robot will move to wait for you to help it up the stairs.
- when requested/needed to move to different floor robot would move to this 'stairs' location
- it would then loop:
  - play a sound to signal it needs help up/down the stairs
  - try to move a bit, if lifted we know someone is helping us
- once we know we are lifted we can assume we are being dropped of at the right location
- loop:
  - try to move: once lifted error is gone we know we are in the new floor
- do the action that was required
- move back to the 'stairs' location in the second floor and again do the 'help me' procedure.

ppisljar:
for initial PR however i think this is fine. rather than above the next improvement would be showing all maps in map view (allowing you to view all saved maps and configure zones for them)

hagenuck1:
Sounds awesome, will build the image and test mqtt tomorrow.

ppisljar:
currently restoring should work:
- save map (Floor1)
- if map is reset load map (Floor1)

it would also be possible to automatically do this if map was reset (set the default floor, on map reset load the default map) but i think that might generate other issues so for now i am leaving it out.

hagenuck1:
Just compiled and tried it. 

Web Interface in Valetudo is working as expected 👍

Mqtt is working fine, too. But errorhandling for non existing floor names is missing and causes the current map to be deleted.

ppisljar:
i will add more error checking

vincegio:
Please add an option in settings to enable/disable these buttons, if this is not done already :)

ppisljar:
this should be ready now

hagenuck1:
Just compiled the new version.

Edit the Multifloor settings and saving it causes valetudo to show "there was an error: 400" to be shown in the bottom.

Store Maps causes it to show an error 500 for me. In the file system it creates an folder with my entered name, but the folder is empty.

Loading maps using the website works fine.

vincegio:
Was curious of the work so booted it up quickly without having my vacuum on the net at the moment.
image
Maybe look into so the buttons are disabled in these states?

hagenuck1:
Thanks, great work! 

Tested and saving & loading map is working fine in web view and mqtt. 

Didn’t found any issues.

ppisljar:
anything to be done before it can be merged ?

hagenuck1:
Didn’t got any feedback and haven’t found more issues, I think it could be merged.

pszafer:
@ppisljar maybe you could add status to mqtt which map/floor is currently loaded.

ppisljar:
@pszafer i can do that in the next PR, first i'll wait for this to get merged.

Hypfer:
I don't trust it. Provide proof that it will work reliably (long term!) or cease.

iMicknl:
@Hypfer is it an idea to have an experimental branch + build for those kind of features?

TheRealMoeder:
I can see the issue you have with this. Maybe an experimental branch with builds will indeed make testing more widespread.

hagenuck1:
I also can fully understand you Hypfer and know that the unknown factory resets should be fixed before getting more potential crashing issues. 

But just to be clear, the RoboController.cfg isn’t needed to be changed for map reloading? Only thing that’s needed to reload the map is restarting the rrwatchdoge to reload the map files. 

For me the loading of the map files using manual scripts and now this pr is working fine since about 2019-06-06 and the robot hadn’t any resets or problems till now. 

Also it survived the 8 week reset and runs fine since 9 weeks now. (Recreated the image to use a new ssh key) 

My map files changing date is 2019-07-07, so they’re working longer as 6/8 weeks as somebody mentioned somewhere.

ppisljar:
i also didn't have any issues so far, the robot works as expected. Also not sure what core functionality am i messing with here :| . i also agree that there should be a development branch and stable one. No idea how i can provide the proof but if this contribution is unwanted i also have other things to do :)

Hypfer:
Well that turned out wrong.
Let me give you the full version of what I was trying to say further up in this thread:

`I don't` think that just looking at the Code to verify that everything is working will be enough here, due to the fact, that there is another Component involved which doesn't have a well-defined interface nor sourcecode available to check the behaviour. If you mention me to get me to merge this PR, please   `provide proof that it will work reliably (long term!) or cease` mentioning me altogether.

I just hate emails like that. Still I should've been more clear there.

Questions/ that would need to be resolved before this can be merged:
- Do we know for certain that simply swapping the data files and restarting the cleaner process is all we need to do to make multiple maps work? Where did that information come from?
- Are we sure that there is no hidden reset counter somewhere which increases each time the process is respawned?
- Does it work when used daily? We got at least one report here from hagenuck1 so thats something
- What about Gen1 vacuums?
- ~~What about the overall code quality~~ ✔️

There's also a strategic decision to be made here: Should Valetudo stick to (semi-) official APIs, or should it use other means to get things done. What about portability to other Vacuums at some point?

hagenuck1:
More info by me: 
Where did the information come from? 
- I’ve read about the last_map replacement in the Valetudo group and figured out by testing, that replace it (NOT during cleaning!) and then restart rrwatchdoge (and not the whole robot as written in the group is enough) 

Hidden counter: 
- can’t provide there’s no hidden counter unfortunately 

Daily usage: 
- no problems for me 
- see the home assistant 0.100 release notes, there’s somebody who has written a tutorial using my wiki fork which is the same as this PR does. He resets the map every time the robot hits docked. 
— https://twitter.com/oischinger/status/1178254181221769216?s=21
- maybe we could let them report issues somewhere if they find problems, too. 

Gen1
- as they don’t have persistent Maps this shouldn’t work. I think for them the multiple maps feature is inactive.

oischinger:
Hi,
I use a gen1 robot. I didn't use the firmware but I restore via ssh as per instructions. The restoring works fine for me.
I was brave enough to trigger the map reset multiple times in a row in a few seconds and didn't see a reset or reboot.

I'll do some more experiments soon on a firmware including this PR.

pax2011:
> Hi,
> I use a gen1 robot. I didn't use the firmware but I restore via ssh as per instructions. The restoring works fine for me.
> I was brave enough to trigger the map reset multiple times in a row in a few seconds and didn't see a reset or reboot.
> 
> I'll do some more experiments soon on a firmware including this PR.

Do you restore maps in which moment? before start cleaning? Do you use homeassistant to restore it?

oischinger:
I set up an automation that resets the map whenever the robot is docked. Because of these reasons:
a) my Gen1 robot sometimes rotates the map by 90deg which brakes zoned cleanup 
b) I have multiple floors and sometimes carry the robot to another floor. The charger is in the ground floor. I want to restore the map of this floor automatically.

But I also triggered the restore automation multiple times in a row (when it was not cleaning). The restore still worked as expected.

pax2011:
> I set up an automation that resets the map whenever the robot is docked. Because of these reasons:
> a) my Gen1 robot sometimes rotates the map by 90deg which brakes zoned cleanup
> b) I have multiple floors and sometimes carry the robot to another floor. The charger is in the ground floor. I want to restore the map of this floor automatically.
> 
> But I also triggered the restore automation multiple times in a row (when it was not cleaning). The restore still worked as expected.

I set up everything, and it works locally on robot. If I try to run as a script on Hass,io i get 

> Error running command: `ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i /config/ssh/id_rsa root@192.168.1.220 "bash ../mnt/data/valetudo/maps/floor1.sh"`, return code: 255
NoneType: None

This is my scipt https://ibb.co/bXHpnwH
This is my config https://it.imgbb.com/

What am I missing?

pax2011:
> I set up an automation that resets the map whenever the robot is docked. Because of these reasons:
> a) my Gen1 robot sometimes rotates the map by 90deg which brakes zoned cleanup
> b) I have multiple floors and sometimes carry the robot to another floor. The charger is in the ground floor. I want to restore the map of this floor automatically.
> 
> But I also triggered the restore automation multiple times in a row (when it was not cleaning). The restore still worked as expected.

Now I would like to add another floor with different zones, zones are stored in  /mnt/data/valetudo/config.json. So I'm trying a script thats overwrite config.json and last_map to get 2 floor with differents zones.
It works, but to have correct zones i have to reboot the vacuum. The file is correct but valetudo's interface shows wrongs zones.

SH scipt
`rsync -r ../../../../mnt/data/valetudo/maps/floor1/ ../../../../mnt/data/rockrobo/
rsync -r ../../../../mnt/data/valetudo/maps/jsonfloor1/ ../../../../mnt/data/valetudo/
rm -f ../../../../mnt/data/rockrobo/user_map0
rm -f ../../../../mnt/data/rockrobo/StartPos.data
service rrwatchdoge restart`
Than reboot.

any solution to reload the config.json without rebooting?

Hypfer:
@pax2011 uh.. restart valetudo?
But that will cause the msgIDs to get out of sync which you also don't want.

The config file is read only once on startup. Then, everything happens in-memory. If changes occur, the whole in-memory config will be written to disk.
This can be seen here:
https://github.com/Hypfer/Valetudo/blob/master/lib/Configuration.js

Since the code is open source after all, I'd suggest working with it instead of around it.

pax2011:
> @pax2011 uh.. restart valetudo?
> But that will cause the msgIDs to get out of sync which you also don't want.
> 
> The config file is read only once on startup. Then, everything happens in-memory. If changes occur, the whole in-memory config will be written to disk.
> This can be seen here:
> https://github.com/Hypfer/Valetudo/blob/master/lib/Configuration.js
> 
> Since the code is open source after all, I'd suggest working with it instead of around it.

thanks for your advice, but I really don't know how to do it. I don't have much experience with coding.

The zones are stored in config.json can't get other solution to reload it.
I also thought about doing this:
`killall miio_client && service valetudo restart`
But it would be the same thing, wouldn't it?

hagenuck1:
> I set up an automation that resets the map whenever the robot is docked. Because of these reasons:
> a) my Gen1 robot sometimes rotates the map by 90deg which brakes zoned cleanup
> b) I have multiple floors and sometimes carry the robot to another floor. The charger is in the ground floor. I want to restore the map of this floor automatically.
> 
> But I also triggered the restore automation multiple times in a row (when it was not cleaning). The restore still worked as expected.

Okay, sounds great, that it's working fine for gen1, too!
I hat the 90deg rotation, too on my gen2 before the persistent maps, I think they're caused by starting a full clean by clicking "start". 
- Solution for this should be to create a zone, that contains the whole floor.
Would be great to hear if the PR is also working fine on gen1.

@pax2011 just create all zones directly in valetudo by loading the map first and add a leading name (for example "floor1_") to each zone containing a hint to the current floor. Then repeat this for the other floor(s). This way you could still differentiate between the rooms and have all zones in one config.json.

jonathan-reisdorf:
Nice! :+1: These "experimental" features could also be put into a separate section in the UI to indicate that it's potentially more dangerous to use them, e.g. adding a new card in settings, or combine it with the persistent data feature ("Map settings"?). (Also one could argue that swapping the floor is something between a setting and an action, so it wouldn't be wrong in settings imo.)

oischinger:
My robot recently did a factory reset but it seems it was not related to the map restoring. It reset itself overnight, around 10 hours after the last map restore.

lib/MapManager.js
```diff
@@ -0,0 +1,75 @@
...
+    if (!fs.existsSync("/mnt/data/rockrobo/last_map")) {
+        return cb("map doesn't exist");
+    }
+    fs.copyFileSync("/mnt/data/rockrobo/last_map", "/mnt/data/valetudo/maps/" + name + "/last_map");
+    if (fs.existsSync("/mnt/data/rockrobo/ChargerPos.data")) fs.copyFileSync("/mnt/data/rockrobo/ChargerPos.data", "/mnt/data/valetudo/maps/" + name + "/ChargerPos.data");
+    else {
+        // we don't have charger info
+    }
+    if (fs.existsSync("/mnt/data/rockrobo/PersistData_1.data")) fs.copyFileSync("/mnt/data/rockrobo/PersistData_1.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_1.data");
+    if (fs.existsSync("/mnt/data/rockrobo/PersistData_2.data")) fs.copyFileSync("/mnt/data/rockrobo/PersistData_1.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_2.data");
```
N1k145: PersistData_1 is copied twice.
Is this correct?

```suggestion
    if (fs.existsSync("/mnt/data/rockrobo/PersistData_2.data")) fs.copyFileSync("/mnt/data/rockrobo/PersistData_2.data", "/mnt/data/valetudo/maps/" + name + "/PersistData_2.data");
```

BoukeHaarsma23:
[Copy from from #91]
Just wanted to add something here. Apparently there is a beta with multi-room support floating around.

See [the announcement on facebook.](https://www.facebook.com/groups/1963580733664463/permalink/2865795586776302/)

If somebody has there hands on it. Maybe it is possible to use this in valetudo as well?

jonathan-reisdorf:
The new firmware 2008 is linked on vacuumz.info and can be downloaded directly from the xiaomi CDN by assembling the base URL `https://cdn.awsbj0.fds.api.mi-img.com/rubys/updpkg/` and the `FileName` displayed on said website for FW 2008.

jonathan-reisdorf:
Didn't test it yet but the announcements on facebook sound promising (also adds automatic room identification and selective room cleaning)

hagenuck1:
It’s only multi room support and not multi floor, as I understood.

jonathan-reisdorf:
Thanks for clarifying. Then it's not related to this PR. (Even though it would be great if somebody tested the new FW and reported if it works with Valetudo, and, if it's still possible to factory reset to original FW to regain rootability)

arieltools:
As I was reading Hypfer's concern [here](https://github.com/Hypfer/Valetudo/pull/317#issuecomment-540929686) about portability to other vacuums, it could also be considered to group these "advanced features" into "add-ons" or "plugins". This will allow to set a "compatibility flag" for each add-on, effectively grouping these per-vacuum in a future multi-vacuum release. 

Then, Valetudo could stick to the [semi-]official APIs and to develop the core further, while allowing for more advanced features to be developed in parallel by the core team or other interested parties.

> Nice!  These "experimental" features could also be put into a separate section in the UI to indicate that it's potentially more dangerous to use them, e.g. adding a new card in settings, or combine it with the persistent data feature ("Map settings"?). (Also one could argue that swapping the floor/map is something between a setting and an action, so it wouldn't be wrong in settings imo. On the other hand, so is changing the fan speed...)

white-gecko:
Why is this closed? Is there a solution for the multi floor issue?

alexkn:
There is no progress here.
#69 is still open, if someone want to implement it in an good way.

PR #335, opened 2019-10-03
Global vars

pszafer:
Hi, so this would be my idea about global vars. It's not finished, but first wanna ask if this is good approach.

One idea which might be good.
JSON like this:
```
"key": {
  "value": "myvalue",
  "httpOK": true/false
}
```

and then create on server side more dynamic global vars and pass to client side only those which have `httpOk == true`. This way it would be easy to hide some of them from client html.


client/home.html
```diff
@@ -62,6 +62,10 @@
         var loadingBarHome = document.getElementById('loading-bar-home');
 
         var config = {};
+        var consts = {}
+        fn.request("api/get_const_vars", "GET", function (err, res) {
```
Hypfer: This won't work this way. JS is asynchronous so you need to wait for this request to finish before continuing everything else, which is most definitely a near-impossible task utilizing the current client. 

I'd suggest adding the constants at "compile time" using a task runner and extending the npm build command definition in the `package.json`

Hypfer: Server-side looks good, though 👍

PR #336, opened 2019-10-04
Fix MQTT autoconf features

TheRealMoeder:
Fixes feature list for MQTT autoconf

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -19,6 +19,7 @@ const CUSTOM_COMMANDS = {
 
 //TODO: since this is also displayed in the UI it should be moved somewhere else
 const FAN_SPEEDS = {
+    whisper: 1,
     min: 38,
     medium: 60,
     high: 75,
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -102,7 +102,7 @@ const MqttClient = function (options) {
                 "start",
                 "pause",
                 "stop",
-                "return_home",
+                "return_to_base",
                 "battery",
                 "status",
                 "locate",


roflcoopter:
Home Assistant doesnt work anymore because of the change to ´´´return_to_base´´´ from ´´´return_home´´´. Shouldn't this PR be reversed or am i missing something?

TheRealMoeder:
No expert on Home Assistant, but from what I read return_home is legacy and return_to_base is the new state scheme (
https://www.home-assistant.io/integrations/vacuum.mqtt/). You should check your HA configuration and probably rediscover the vacuum device

Hypfer:
https://www.home-assistant.io/integrations/vacuum.mqtt/#legacy-basic-commands

`return_to_base` is legacy..

TheRealMoeder:
Are you sure? From that page:

`State MQTT Protocol

The above configuration for this integration expects an MQTT protocol like the following.

STATE BASIC COMMANDS

MQTT topic: vacuum/command

Possible MQTT payloads:

start - Start cleaning
pause - Pause cleaning
return_to_base - Return to base/dock`

roflcoopter:
I'm pretty sure. If you update to the latest home assistant version and rediscover the vacuum, it doesn't work with this PR.

TheRealMoeder:
there seems to be an error in the HA documentation then, because the description differs from the configuration.yaml example for the State MQTT Protocol

TheRealMoeder:
After reading the HA code the that is probably where the mess begins. It expects a feature „return_home“ in auto-discovery but actually sends a state „return_to_base“ via Mqtt. I wonder of this is intended...

PR #340, opened 2019-10-08
MQTT: Add timers

MarBra:
I like to provide the list of timers via MQTT. 
So we can use it for example in Homeassistant to display the configured timers.

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -263,6 +263,15 @@ MqttClient.prototype.updateAttributesTopic = function () {
                             response.zoneStatus = zoneCleaningStatus
                         }
 
+                        this.vacuum.getTimers( (error, timer_response) => {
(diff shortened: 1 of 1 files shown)


Hypfer:
This will not work due to the fact that the getTimers call is asynchronous, but the rest of the code is not waiting for it to return.

It might have worked on the second update but that's just because you've read stale data which was saved to the this context somewhere soon after the previous response.

MarBra:
> This will not work as expected due to the fact that the getTimers call is asynchronous, but the rest of the code is not waiting for it to return.
> 
> It might have worked on the second update but that's just because you've read stale data which was saved to the this context somewhere soon after the previous response.

Sorry i'm not really familiar with nodejs :see_no_evil: 
Should i store the timers first in a property like we do it for zoneCleaningStatus?

PR #360, opened 2019-11-01
refactor: Convert issue template bolds to headers

x10an14:
So as to not break on `**Valetudo version: **` e.g.

Also group `Vacuum Model`, `Version`, `Screenshots`, and `User-Agent`
under `To reproduce`.


x10an14:
Maybe also switch order of `Reproduce` and `Expected behaviour`? 

Nitpicking, though I think it makes more sense to start off with what you desire, then spend (a potential lots of) text to describe reproducibility.

PR #369, opened 2019-11-09
Make action sheets scrollable

TheLastProject:
Fixes the issue MichaelBW ran into and reported on Telegram: https://t.me/c/1357472831/10875

The minimized CSS files still need to be updated though, but I'm not sure how to best do so.


florian-asche:
looks ok to me.

PR #371, opened 2019-11-10
add telegrambot version 1

thejoe8495:


Hypfer:
I think this would be better suited as an external component which talks to the robot via MQTT

thejoe8495:
But i need the dummycloud to get the info when the robot starts over the Timer. In this case i need a dummycloud, when i need the dummycloud i can only install Telegrambot or the webinterface.

Hypfer:
I see. Those events should also be pushed to mqtt then 🤔

thejoe8495:
Or anyone create a service dummycloud that listen on a Port and send it over this Port to both webinterface and telegram. But this is many work to find a good protocol to send the infos.

leonardpitzu:
I've done something similar since day one - i have a simple bash script polling the API for changes. I've cron-ed the script to run once a minute so i get the status changes with some second of delay which is acceptable for me. The script also serves as a watchdog - if Valetudo hangs it restarts it, if it's "dead" it kills off the process and then restarts it.

PR #373, opened 2019-11-13
adds proper shutdown on SIGTERM and SIGINT

bbarbisch:
Shutdown handler on unix signals:
- SIGTERM
- SIGINT

closes all open sockets, clients and shuts down the webserver before valetudo is closed.

Resource cleanup and closing open file descriptors (sockets, files,...) is always a good idea when shutting down an application (sometimes file systems can get corrupt when they could not be unmounted correctly during os shutdown due to  e.g. open file descriptors) . So this change might help with the reset issue but I'm not really sure about that.

Please review especially the listed points, as I'm petty new to the code base and node/javascript in general:
- I think all used calls for closing resources are synchronous calls. If not we/I need to refactor the synchronous shutdown methods to asynchronous methods.
- In some classes files are opened for read/write (e.g. Configuration.js). As far as I understand the docs these calls close the opened files automatically?
- Webserver.js:947: had to add this as a workaround (it seems that the variable 'server' needs to be const as it us used later on in a callback) to get to the webserver instance. Maybe someone has a better idea?

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -416,13 +416,12 @@ MqttClient.prototype.handleCustomCommand = function (message) {
     }
 };
 
-MqttClient.prototype.shutdown = function() {
-    console.debug("Shutting down the MQTT Client...")
-    
-    //force close the mqtt client
-    this.client.end(true);
-
-    console.debug("Shutting down the MQTT Client done.")
+MqttClient.prototype.shutdown = function(callback) {
+    console.debug("Shutting down the MQTT Client...");
+    this.client.end(true, () => {
+        console.debug("Shutting down the MQTT Client done");
+        callback();
+    });
 }
 
 module.exports = MqttClient;
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -416,4 +416,13 @@ MqttClient.prototype.handleCustomCommand = function (message) {
     }
 };
 
+MqttClient.prototype.shutdown = function() {
+    console.debug("Shutting down the MQTT Client...")
+    
+    //force close the mqtt client
+    this.client.end(true);
+
+    console.debug("Shutting down the MQTT Client done.")
+}
+
 module.exports = MqttClient;
--- a/lib/Valetudo.js
+++ b/lib/Valetudo.js
@@ -58,6 +58,7 @@ const Valetudo = function() {
 
 
     if(this.configuration.get("mqtt") && this.configuration.get("mqtt").enabled === true) {
+        console.log("mqtt start");
         this.mqttClient = new MqttClient({
             configuration: this.configuration,
             vacuum: this.vacuum,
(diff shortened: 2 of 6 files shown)


lib/Valetudo.js
```diff
@@ -132,4 +132,24 @@ Valetudo.VACUUM_MODEL_PROVIDER = function() {
     }
 };
 
+Valetudo.prototype.shutdown = function() {
+    console.info("Valetudo shutdown in progress...")
+    
+    //shutdwon mqtt client
+    if(this.configuration.get("mqtt") && this.configuration.get("mqtt").enabled === true) {
```
Hypfer: Instead of checking this, you could just call it every time and extend the MqttClient shutdown method to only act if there really is a connection

lib/MqttClient.js
```diff
@@ -416,4 +416,13 @@ MqttClient.prototype.handleCustomCommand = function (message) {
     }
 };
 
+MqttClient.prototype.shutdown = function() {
+    console.debug("Shutting down the MQTT Client...")
+    
+    //force close the mqtt client
+    this.client.end(true);
```
Hypfer: https://github.com/mqttjs/MQTT.js#end

This can accept a callback which we might want to wait for

lib/Valetudo.js
```diff
@@ -132,4 +132,24 @@ Valetudo.VACUUM_MODEL_PROVIDER = function() {
...
+    //shutdown webserver
+    this.webserver.shutdown();
+    
+    //shutdown vacuum
+    this.vacuum.shutdown();
+    
+    //shutdown dummycloud
+    this.dummycloud.shutdown();
+
+    console.info("Valetudo shutdown done.")
```
Hypfer: Since (some of) those operations above are asynchronous, this line might appear even if they aren't shut down yet.

The async library contains some useful functions to avoid the callback hell for such applications:
https://caolan.github.io/async/v3/docs.html#parallel

I'm still looking for nice ways to achieve the same tasks without additional libraries so if someone knows something please comment. 🙃

lib/Valetudo.js
bbarbisch: A check is necessary because the mqttClient instance is only available if it is configured in the config. If I would simply call this.mqttClient every time I would get a NullPointer/NullReference (or whatever the javascript error is) every time, when no mqtt is used.
To make things easier it can also be checked by `if(this.mqqtClient)...` (changed this in the second commit).

lib/MqttClient.js
bbarbisch: yep. added it in the second commit

lib/Valetudo.js
bbarbisch: Refactored the calls to a async calls.

The current solution is based on callbacks, but it also introduces the thing called 'callback hell'.

I played around with Promises and async/await during the last couple of days but wasn't successful. Especially the handling of inner callbacks (mqttclient.end) with promises gave me some headaches. Maybe nodes util.promisify can help, but I'm not sure.

Issue #368, opened 2019-11-09
[Feature Request] Dark Theme


silviob72:
Dark Theme for WebUI. Also with time for turn on/off automatically and with sunrise sunset, or via mqtt command.

pax2011:
It would be nice!!

rimrul:
I've done rough quick dark theme toggle with browser developer tools and the dark css that already ship with valetudo:

![dark home](https://user-images.githubusercontent.com/6178234/68958906-52f99580-07cd-11ea-9917-a25e2cfd44bf.png)
![dark settings](https://user-images.githubusercontent.com/6178234/68958921-5db42a80-07cd-11ea-9fdd-9e04f49c3bab.png)

There's still some improvements to be made...

silviob72:
Cool! 👍

pax2011:
On chrome I use [https://chrome.google.com/webstore/detail/dark-reader/eimadpbcbfnmbkopoojfekhnkhdbieeh](url) but only on pc
![val](https://user-images.githubusercontent.com/56022493/68964143-ebe1de00-07d8-11ea-94f5-5e6c32cc899c.PNG)

notz:
It would be cool the to load own css theme per browser. So we can override the some styles if you embed valetudo in home assistant as iframe.

erkexzcx:
Someone was doing God's work by bringing dark theme.

rimrul:
New and improved state of #374 :

![grafik](https://user-images.githubusercontent.com/6178234/69370593-88572500-0c9e-11ea-9079-4fc0f7fea99f.png)
![grafik](https://user-images.githubusercontent.com/6178234/69370653-a755b700-0c9e-11ea-92a1-ac28e5beef2b.png)

Hypfer:
PR is already merged to master. Thanks @rimrul

PR #374, opened 2019-11-15
Dark mode

rimrul:
This should, if it works add the rudimentary dark mode toggle I mentioned in #368.
I still need to test this, though.


rimrul:
I've updated the dark css using the [OnsenUI theme roller](https://onsen.io/theme-roller/?customize) and unintentionally semi-reverted the changes from 4ba4468. I'll readd those and squash the css commits into one. 

Do we also need to apply the 4ba4468 changes to `dark-onsen-css-components.min.css` and `onsen-css-components.min.css` for that change to actually work as intended?

rimrul:
Apparently the theme roller generated a broken light theme instead. Not exactly what I was attempting to do.

rimrul:
The sidebar switch works, but I'm not currently modifying the map colours and the theme selection does not survive a refresh.

It's a start, but there's room for improvement.

Hypfer:
What about persisting the theme settings to local storage so that they survive a refresh?

Should be easy to do

rimrul:
Yes, some kind of persistence would definitly be good. I'm just not sure what the best approach to this is.

#368 suggests an MQTT command (which implies a setting stored on the vacuum) and sunset/sunrise timers (I guess based on vacuum time, but could be based on client time).

But what about multiple users (client browsers) with different preferences? Maybe cookies/local storage/indexed db are a better choice.

There's also `prefers-color-scheme` media queries, which I'd consider the proper thing to respect, but those [don't play nicely with `privacy.resistFingerprinting` in Firefox.](https://bugzilla.mozilla.org/show_bug.cgi?id=1535189)

rimrul:
Ok, I've thought about it some more and decided to go with local storage and 'prefers-color-scheme' media queries. I've decided the explicit choice for Valetudo overrides the more general media queries. 

I'll leave MQTT and timers to others in the future.

Issue #337, opened 2019-10-07
Hassio no vacuum.valetudo over mqtt


dennis071nl:
Is there a way to manually trigger the connect message??

I cannot operate the vacuum through HomeAssistant (Hassio 0.99.3)

I followed the installation guide, succesfully. I get the frontend, can start/stop cycles etc etc. I ssh-ed into the rockrobo, entered my MQTT details and rebooted. In the log says MQTT server connected, only thing is that the integration doesn't work . There is no vacuum.rockrobo in my entities. 

**To Reproduce**

**Expected behavior**

**Screenshots**

Logfile in Valetudo says  "
Robot connected
Robot connected
Robot connected
Robot connected
Loading configuration file: /mnt/data/valetudo/config.json
Dummycloud is spoofing 203.0.113.1:8053 on 127.0.0.1:8053
Webserver running on port 80
Connected successfully to mqtt server
Robot connected
Robot connected

In Home Assistant the log of the MQTT addon says:

INFO] found mqtt on Home Assistant
1570460454: New client connected from 192.168.2.21 as mqttjs_22bbbc61 (p2, c1, k60, u'mqtt').

**Vacuum Model:** 
Gen 1
**Valetudo Version: ** 
0.4.0
**User-Agent** 

**Additional context**

oechslein:
Do you have an entry in your logfile with
ERROR (MainThread) [homeassistant.components.mqtt.vacuum] Exception in async_discover when dispatching 'mqtt_discovery_new_vacuum_mqtt': ({'name': 'rockrobo', 'schema': 'state', 'supported_features': ['start', 'pause', 'stop', 'return_to_base', 'battery', 'status', 'locate', 'clean_spot', 'fan_speed', 'send_command'], 'command_topic': 'valetudo/rockrobo/command', 'state_topic': 'valetudo/rockrobo/state', 'set_fan_speed_topic': 'valetudo/rockrobo/set_fan_speed', 'fan_speed_list': ['whisper', 'min', 'medium', 'high', 'max', 'mop'], 'send_command_topic': 'valetudo/rockrobo/custom_command', 'json_attributes_topic': 'valetudo/rockrobo/attributes', 'platform': 'mqtt'},)
?

I supect there is currently a mismatch between the autoconfig message what valetudo sends ([after this change](https://github.com/Hypfer/Valetudo/commit/92d7285b2a3bf76096ce86fb5a7f823f133960b8)) and what home assistant expects (v0.100.2).
[...]

PR #377, opened 2019-11-18
Add unique_id for MQTT Discovery

MitschiBoy:
See https://www.home-assistant.io/docs/mqtt/discovery/ and https://github.com/home-assistant/home-assistant/pull/20503
Without unique_id MQTT Discovery isn't working with actual versions of Home Assistant


MitschiBoy:
This should fix #337

Issue #426, opened 2020-02-12
Modernize Infrastructure: Support ES6 / Typescript


rumpeltux:
imho it would be beneficial for the project to modernize the infrastructure so that new code can be written in typescript or at least ES6.

This has a bunch of benefits like:
* having proper support in IDEs, whose automated error checking can avoid a whole class of bugs
* being able to more easily define interfaces and class hierarchies
* potentially being able to compile the code to a more efficient representation which may alleviate some of the binary-size concerns

Similarly the project should probably adopt a linter. I’m not sure what the state of the art is for either though.

rumpeltux:
That is assuming that the new architecture is still based on nodejs (which I think does make sense given the existing codebase and the easiness of the development cycle).

alexkn:
I am unsure whether we should use Typescript or ES6.
@rumpeltux  What are the advantages of Typescript compared to ES6 with activated type checking?
If there are no real advantages, I would want to avoid the additional compilation of Typescript.

I use ESLint in some other projects.
I also created GitHub Actions there to automatically test commits and pull requests with the Linter.

rumpeltux:
I found https://stxnext.com/blog/2019/08/30/typescript-pros-cons-javascript/
tl;dr benefits are minimal once you consequently use jsdoc annotations, examples less verbose type writing, compile-time errors

So sticking with jsdoc makes sense. ES6 classes seem to work. ES6 imports don't, but I guess I can live with that.

@alexkn Could you set up eslint for the project?

alexkn:
> @alexkn Could you set up eslint for the project?
Yes, I will do that.

alexkn:
@rumpeltux Could you refactor the ViomiMapParser to a class?
Then we can close this issue.

rumpeltux:
> @rumpeltux Could you refactor the ViomiMapParser to a class?
> Then we can close this issue.

Done :-)

PR #427, opened 2020-02-12
Add ESLint

alexkn:
as mentioned in #426 

including initial fixes
and GitHub Action to automatically lint commits

You can use the Visual Studio Code Extension "ESLint"
or you can run the linter manually with:
`npm run lint`

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,8 @@
+// ignore external libraries
+/client/js/onsenui.js
+/client/js/onsenui.min.js
+/client/js/pako_inflate.min.js
+/client/js/angular-onsenui.js
+/client/js/angular-onsenui.min.js
+/client/js/geometry-polyfill.js
+/client/zone/geometry-polyfill.js
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,6 +19,10 @@
         "linebreak-style": [
             "error",
             "unix"
+        ],
+        "semi": [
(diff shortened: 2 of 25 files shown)


rimrul:
Would it make sense to combine the CI job with the existing Travis CI? Looking at the Travis job it just seems to run 'npm build' in trusty. Github doesn't seem to offer trusty runners. So I guess if we merge them we'd need to merge them on trusty.

alexkn:
I also wondered if I should merge that.
But as you've already noticed, there are no trusty runners on GitHub.

I am not sure if that job really has to run on trusty.
The build even works on windows.

I have never used Travis.
GitHub actions have the advantage that they also run in forks without any further configuration.

I would let the jobs run parallel for now.

rimrul:
Ok, sounds like a solid plan. Testing on Trusty probably makes more sense than just building on trusty. I'll investigate if we could get a trusty-armv7 docker image going on github (using [uraimo/run-on-arch-action](https://github.com/uraimo/run-on-arch-action) and a custom image) for when we have some tests. I don't think we'll get a tina linux container for Viomi CI, though.

rumpeltux:
Tina Linux isn’t really required. I think the core params are armv6 or armv7 with musl libc. I believe the latter is covered by images labeled alpine.

alexkn:
You should open a new issue for testing ideas.

PR #430, opened 2020-02-15
Add logger

### LINKED ISSUE #154: Improve Logging
Valetudo needs configurable loglevels instead of randomly added/removed console.log statements.

alexkn:
It adds timestamps and the log level to the messages and then redirects them to the console.
Loglevel can be configured in valetudo config.json

closes #154

### DIFF
--- a/lib/Logger.js
+++ b/lib/Logger.js
@@ -21,6 +21,9 @@ class Logger {
         return Object.keys(LogLevels).find(key => LogLevels[key] === LogLevel);
     }
     static set LogLevel(value) {
+        if(!LogLevels[value]) {
+            throw "invalid log level '" + value + "'";
+        }
         LogLevel = LogLevels[value];
     }
 
--- a/lib/Logger.js
+++ b/lib/Logger.js
@@ -1,0 +1,83 @@
+let LogLevel = 0;
+
+const LogLevels = {
+    trace: -2,
+    debug: -1,
+    info: 0,
+    warn: 1,
+    error: 2
+};
+
+const BuildPrefix = (LogLevel) => {
+    let timestamp = new Date().toISOString();
+    return "[" + timestamp + "] [" + LogLevel + "] ";
+};
+
+class Logger {
+    /**
+     * @property {"trace"|"debug"|"info"|"warn"|"error"}
+     */
+    static get LogLevel() {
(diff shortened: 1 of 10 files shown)


rumpeltux:
Would it make sense to also log the originating filename?

Not sure if there’s a better way, but for example using:

    function getCaller() { return new Error().stack.split('\n')[2].substring(7); }

alexkn:
I don't think this is necessary.
The log message should provide enough information about where the error came from.
If you need that to find the calling function, then the log message in the originating file is too bad.

lib/Logger.js
```diff
@@ -0,0 +1,86 @@
...
+     */
+    static warn(message, ...args) {
+        if(LogLevel <= LogLevels.warn) {
+            console.warn(BuildPrefix("WARN") + message, ...args);
+        }
+    }
+
+    /**
+     * @see console.error
+     * @param {string} message
```
rumpeltux: This is often used with an Error instance as the first argument.
Should the signature in general be just `...args` (i.e. without `message`)?

lib/Logger.js
```diff
@@ -0,0 +1,86 @@
+let LogLevel = 0;
+
+const LogLevels = {
+    trace: -2,
+    debug: -1,
+    info: 0,
+    warn: 1,
+    error: 2
+};
```
rumpeltux: You can wrap this in Object.freeze since there are no plans for extensions.

The keys should be quoted since they are used as strings elsewhere (in minification non-quoted keys may get renamed).

lib/Logger.js
```diff
@@ -0,0 +1,86 @@
...
+class Logger {
+    /**
+     * @property {"trace"|"debug"|"info"|"warn"|"error"}
+     */
+    static get LogLevel() {
+        return Object.keys(LogLevels).find(key => LogLevels[key] === LogLevel);
+    }
+    static set LogLevel(value) {
+        if(!LogLevels[value]) {
+            throw "invalid log level '" + value + "'";
```
rumpeltux: This should also output which levels are valid. To make the error easier to fix for users themselves.

lib/Logger.js
Hypfer: It does make sense to have one defining message plus corresponding metadata. Otherwise this might get confusing.

I think we can enforce always adding a message to it

PR #433, opened 2020-02-15
refactor to ES6 classes

alexkn:
as proposed in #426 

I left out the miio files because the viomi branch will probably make them obsolete anyway.

### DIFF
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -2,86 +2,59 @@ const fs = require("fs");
 const Tools = require("./Tools");
 const path = require("path");
 const Logger = require("./Logger");
+const DEFAULT_SETTINGS = require("./res/default_settings");
 
-/**
- * @constructor
- */
-const Configuration = function() {
-    this.location = process.env.VALETUDO_CONFIG ? process.env.VALETUDO_CONFIG : "/mnt/data/valetudo/config.json";
-    this.settings = {
-        "spots": [],
-        "areas": [],
-        "mqtt" : {
-            enabled: false,
-            identifier: "rockrobo",
-            topicPrefix: "valetudo",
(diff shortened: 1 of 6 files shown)


lib/Valetudo.js
```diff
@@ -10,71 +10,6 @@ const EventEmitter = require("events");
 const SSHManager = require("./SSHManager");
 const Logger = require("./Logger");
 
-const Valetudo = function() {
```
rumpeltux: Could you leave out this file for now as well? This will create a bunch of merge conflicts, e.g. viomi refactored the readfile stuff.

alexkn:
That changes have nothing to do with viomi.
I don't see why this should be done in the viomi branch.

rumpeltux:
> That changes have nothing to do with viomi.
> I don't see why this should be done in the viomi branch.

Because it’s ahead of master by xx commits and doing concurrent development in diverging branches is creating merge conflicts that need to be resolved which imho unnecessarily costs time. If you think of the `viomi` branch as a `devel` branch putting the changes there doesn’t seem off to me.

lib/Valetudo.js
alexkn: In my opinion you put too much miio specific code in Valetudo.js, without which you wouldn't have the problem.

rumpeltux: You’re right there’s a lot of stuff here that should go into MiioVacuum. Then this should only init MiioVacuum and setup the webserver & mqtt.
However, if I had continued the refactoring to include this file the problem with merge conflicts would remain.

alexkn: Anyway, I took the file out.

rumpeltux:
LG, Thanks.

PR #437, opened 2020-02-18
[viomi branch] refactor to ES6 classes

alexkn:
roborock/viomi classes as mentioned in #433

I tested only the roborock part.
@rumpeltux please test the viomi part.

### DIFF
--- a/lib/devices/MiioVacuum.js
+++ b/lib/devices/MiioVacuum.js
@@ -11,15 +11,15 @@ const Status = require("../miio/Status");
  * Yet incomplete. Only the Roborock implementation is authoritative atm.
  */
 class MiioVacuum {
-    /* @param options {object}
+    /**
+    * @constructor
+    * @param options {object}
     * @param options.cloudKey {Buffer} The pre-shared unique key of your robot
     * @param options.configuration {import('../Configuration')}
     * @param options.deviceId {string} The unique Device-id of your robot
     * @param options.events {import("events").EventEmitter}
     * @param options.ip {string} Remote IP
     * @param options.tokenProvider {()=>Buffer}
-    *
-    * @constructor
     */
     constructor(options) {
         this.status = new Status();
--- a/lib/devices/MiioVacuum.js
+++ b/lib/devices/MiioVacuum.js
@@ -9,169 +9,177 @@ const Status = require("../miio/Status");
  * Common Base class for devices that can be controled by Valetudo.
  *
  * Yet incomplete. Only the Roborock implementation is authoritative atm.
- *
- * @param options {object}
- * @param options.cloudKey {Buffer} The pre-shared unique key of your robot
- * @param options.configuration {import('../Configuration')}
- * @param options.deviceId {string} The unique Device-id of your robot
- * @param options.events {import("events").EventEmitter}
- * @param options.ip {string} Remote IP
- * @param options.tokenProvider {()=>Buffer}
- *
- * @constructor
  */
-const MiioVacuum = function(options) {
-    this.status = new Status();
-    this.events = options.events;
-    this.configuration = options.configuration;
-    this.mapUploadHost = this.configuration.get("map_upload_host");
+class MiioVacuum {
+    /* @param options {object}
+    * @param options.cloudKey {Buffer} The pre-shared unique key of your robot
+    * @param options.configuration {import('../Configuration')}
+    * @param options.deviceId {string} The unique Device-id of your robot
+    * @param options.events {import("events").EventEmitter}
+    * @param options.ip {string} Remote IP
+    * @param options.tokenProvider {()=>Buffer}
+    *
+    * @constructor
+    */
+    constructor(options) {
+        this.status = new Status();
+        this.events = options.events;
+        this.configuration = options.configuration;
+        this.mapUploadHost = this.configuration.get("map_upload_host");
 
-    this.dummycloud = new Dummycloud({
-        spoofedIP: this.configuration.get("dummycloud").spoofedIP,
-        cloudKey: options.cloudKey,
-        deviceId: options.deviceId,
-        bindIP: this.configuration.get("dummycloud").bindIP,
-        onConnected: () => this.onCloudConnected(),
-        onMessage: msg => this.onMessage(msg)
-    });
+        this.dummycloud = new Dummycloud({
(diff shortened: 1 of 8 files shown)


lib/devices/Viomi.js
```diff
@@ -80,85 +162,4 @@ const ERROR_MAP = Object.freeze({
...
-    });
-};
-
-Viomi.prototype.getCurrentStatus = function() {
-    return Promise.resolve(this.status);
-};
-
-Viomi.prototype.pollMap = function() {
-    clearTimeout(this.pollMapTask);
-    this.sendLocalWithPromise("set_uploadmap", [2], {timeout: 2000}).finally(() => {
```
rumpeltux: Hmm sendLocalWithPromise should have these options too, but then we also need to support this for the cloudSocket… I can do that another time though.

lib/miio/Status.js
```diff
@@ -4,16 +4,17 @@
  */
 const Status = function() {
     /** List of allowed values as per Roborock.STATUS_MAP */
-    this.state = "UNKNOWN";
+    this.state = 0;
```
rumpeltux: I think using strings or an enum here is strictly better. And roborock updateStatus also sets this to a string.

lib/miio/Status.js
```diff
@@ -4,16 +4,17 @@
...
     this.error_code = null;
     this.battery = null;
     this.clean_time = null;   // Seconds?
     this.clean_area = null;   // mm²
     this.error_code = null;   // TODO
     this.map_present = false; // TODO
-    this.in_cleaning = false; // TODO
+    this.in_cleaning = 0; // TODO
```
rumpeltux: Why int?

lib/miio/Status.js
```diff
@@ -4,16 +4,17 @@
...
     this.fan_power = null;
     this.dnd_enabled = false;
+    this.lab_status = 0;
```
rumpeltux: If adding a new field here, you should precisely document what it means.

lib/miio/Status.js
alexkn: That was only a fix due to type checking.
There is Code which uses it as number.

lib/webserver/WebServer.js
```diff
@@ -563,7 +563,7 @@ const WebServer = function (options) {
                 ...req.body.virtual_walls.map(wall => [1, ...wall])
             ];
 
-            if (self.vacuum.model === "roborock.vacuum.s5") { //TODO: Move these magic strings to somewhere else
+            if (self.model.RoborockApi) {
```
Hypfer: Finally

lib/miio/Status.js
alexkn: That was only a fix due to type checking.
There is Code which uses it as number.

lib/miio/Status.js
rumpeltux: The one in webserver.js right? Should fix those locations instead.

lib/miio/Status.js
alexkn: That was also a fix due to type checking.
There is Code which uses it as number, but I don't know what each number means.

lib/devices/MiioVacuum.js
```diff
@@ -9,169 +9,177 @@ const Status = require("../miio/Status");
...
-        doTimesync: false
-    });
-}
-
-MiioVacuum.FAST_MAP_PULLING_STATES = [
-    "REMOTE_CONTROL_ACTIVE", "CLEANING", "RETURNING_HOME", "MANUAL_MODE", "SPOT_CLEANING",
-    "DOCKING", "GOING_TO_TARGET", "ZONED_CLEANING"
-];
+class MiioVacuum {
+    /* @param options {object}
```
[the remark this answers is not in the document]
alexkn: Yes, I missed it, either you fix it yourself or I am doing it this evening.

lib/miio/Status.js
rimrul: Looking at [client/settings-persistent-data.html#L102-L167](https://github.com/Hypfer/Valetudo/blob/master/client/settings-persistent-data.html), it seems like `1` means `persistent data enabled` and every other value means `persistent data disabled`.

lib/miio/Model.js
```diff
@@ -1,17 +1,20 @@
...
-    this.name = name;
-};
+class Model {
+    /**
+     * @param {string} name
+     * @constructor
+     */
+    constructor(name) {
+        this.viomiApi = name == "viomi.vacuum.v7";
+        this.RoborockApi = name == "roborock.vacuum.s5";
```
rimrul: This works for what you use it for, but doesn't really indicate what the naming implies. The naming implies that for a Roborock vacuum the property `RoborockApi` is true, while for a Viomi vacuum the property `viomiApi` is true. Instead the property `viomiApi` is true for all currently supported Viomi vacuums and false for all currently supported Roborock vacuums. The `RoborockApi` property does not properly indicate that the `Roborock` vacuum API should be used, but that the robot is a Gen 2 Roborock vacuum and supports persistent data. 

Thinking about future supported devices, maybe we shouldn't set booleans here to define what implementation of `MiioVacuum` to use, but rather set a single property that already references the correct implementation to use.

alexkn: This should be refactored later, it's not the topic of this pull request.

rimrul: Fair enough.

lib/miio/Status.js
alexkn: It is used almost everywhere as a number, even your fakerobot.js sets it to 2.
Maybe it should be refactored later, it's not the topic of this pull request.

lib/miio/Status.js
alexkn: It is used everywhere as a number.
Maybe it should be refactored later, it's not the topic of this pull request.

alexkn:
In my opinion, this can be merged.

lib/miio/Status.js
rumpeltux: fakerobot sends a json object not a Status one.
It is converted by updateStatus to a string: https://github.com/Hypfer/Valetudo/blob/viomi/lib/devices/Roborock.js#L173
The remaining usage in webserver.js as an int is a bug that should be fixed.
But not necessarily here. Feel free to merge and we can fix it later.

PR #440, opened 2020-02-19
Extract javascript of UI page to separate file.

rumpeltux:
This allows us vscode to type-check and xref the file.

I did this exemplarily for one file only for now.
Note I couldn’t get this to work with a normal `<script src="…">` tag. I believe onsen is stripping the `src` attribute. So I write the script tag via javascript.
If anyone knows a better alternative, I’m all ears.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -143,14 +143,16 @@
             return ('0' + h).slice(-2) + ":" + ('0' + m).slice(-2) + ":" + ('0' + s).slice(-2);
         };
         window.fn.applyTheme= function(){
+            let themeEl = document.getElementById('theme');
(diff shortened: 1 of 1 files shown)


alexkn:
Do we want to use a framework like Vue or Angular in the client? 
Type Checking should also be supported there by vs code.

rumpeltux:
I don’t mind a modernization of the UI layer, not sure though which is the right technology for this.
However, I think this would require a much more work-intensive rewrite of the UI. Do you have plans to do so?
I would otherwise argue to focus on the backend parts first.
Any other reasons you’re opposing this?

alexkn:
No, I currently don't have such plans.
I would also focus on the backend first.

PR #442, opened 2020-02-22
Extract javascript of UI page to separate files

rumpeltux:
I accidentally closed the original PR #440.
In contrast to the original, this one extracts js code for all the client/*.html files and fixes eslint errors.
It doesn’t fix all typing errors reported by vscode.

> This allows us vscode to type-check and xref the file.
> 
> I did this exemplarily for one file only for now.
> Note I couldn’t get this to work with a normal `<script src="…">` tag. I believe onsen is stripping the `src` attribute. So I write the script tag via javascript.
> If anyone knows a better alternative, I’m all ears.


alexkn:
The global.d.ts is not loaded automatically, so I always have a lot of type errors.
Do I miss something?
Or is a
/// <reference path="../global.d.ts"/>
missing in each file?

env browser is already set in .eslintrc
So /*eslint-env browser*/ should not be necessary.

rumpeltux:
Ah thanks, removed eslint-env lines.

`global.d.ts` is loaded for me automatically. I get many more errors when I uncomment it. I have set the `checkJs` option to true. I also have a `jsconfig.json` which seems to be a prerequisite for this to work.

alexkn:
I have the checkJs enabled globally in vs code, so I dont't have a jsconfig.json.
I think you should also commit the jsconfig.json.

rumpeltux:
> I have the checkJs enabled globally in vs code, so I dont't have a jsconfig.json.
> I think you should also commit the jsconfig.json.

Done.

alexkn:
Looks good now.

PR #444, opened 2020-02-23
Add initial support for Viomi devices

Hypfer:
One giant leap for valetudo 🌑


rumpeltux:
Woohoo 🎉 Thanks for your help with this @Hypfer and @alexkn!

PR #445, opened 2020-02-24
fix typechecking errors

alexkn:


lib/webserver/WebServer.js
```diff
@@ -906,10 +907,13 @@ const WebServer = function (options) {
             //However for now, it should be sufficient to just limit the concurrent responses per client to one
             if (!err)
                 wss.clients.forEach(function each(ws) {
+                    // @ts-ignore
```
rumpeltux: Another possible workaround is to use ws['mapUploadInProgress'].
Or potentially define the property in global.d.ts (haven’t tested if that works).

alexkn: I also thought about both.

I didn't find a working solution to overwrite/extend the type definition.
If you know one, please do so.

The ws['mapUploadInProgress'] Syntax is just a hackish workaround, then its better to disable the check.

rumpeltux: Can’t get the `global.d.ts` to work either.

The `ws['mapUploadInProgress']` syntax is imho more than a hackish workaround: The type-checkers work on explicitly named properties. Under the hood pretty much every js object is a map though, so map access is always legit. I’d argue it’s preferable, but that’s just my opinion I guess ;-) so feel free to ignore me.

alexkn: I'd rather leave the ts-ignore in there. Maybe someday someone will have an idea how to get the typechecking working.

PR #448, opened 2020-03-02
refactoring with promises/async/await

alexkn:


lib/devices/Viomi.js
```diff
@@ -90,15 +88,23 @@ class Viomi extends MiioVacuum {
...
     }
 
-    preprocessMap(data, callback) {
-        return zlib.inflate(data, callback);
+    preprocessMap(data) {
+        return new Promise((resolve, reject) => {
+            zlib.inflate(data, (err, result) => {
+                if(err) {
+                    reject(err);
+                }else{
```
rumpeltux: (nit-picking) there should be spaces here after `if` and around `else`
could also write this less verbosely with `err ? reject(err) : resolve(result)`

lib/webserver/WebServer.js
```diff
@@ -233,578 +223,548 @@ const WebServer = function (options) {
...
+            if (req.body && req.body.new_zone) {
+                let data = await this.vacuum.setTimezone(req.body.new_zone);
                 res.json(data);
+            } else {
+                res.status(400).send("invalid request");
             }
-        });
+        } catch (err) {
+            res.status(500).send(err.toString());
+        }
```
rumpeltux: This is such a frequent pattern.
Could we use a middleware for this (e.g. as in https://expressjs.com/de/guide/error-handling.html) and avoid all the try-catch blocks alltogether?

lib/devices/Viomi.js
alexkn: fixed and added eslint keyword-spacing rule

lib/webserver/WebServer.js
alexkn: Good idea, but it seems that it doesn't work with the async functions.
I would leave it that way for now.
If I find a way that works, I send a new PR.

alexkn:
Why it's not getting merged?

PR #453, opened 2020-03-07
Actually poll status in pollStatusLoop (broken by 4f8409c)

rumpeltux:
@alexkn

### DIFF
--- a/lib/devices/MiioVacuum.js
+++ b/lib/devices/MiioVacuum.js
@@ -58,13 +58,13 @@ class MiioVacuum {
             options.tokenProvider
         );
 
-        // Poll status once per minute.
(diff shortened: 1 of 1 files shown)


alexkn:
oops
But strange that I didn't see any negative effects in my tests.

rumpeltux:
That’s probably because for Roborock, the web action for querying the status, queries and updates the status on the robot. I assume that happens frequently enough to bring the map-polling in sync fast enough.
For Viomi, Valetudo polls the status and then just returns the currently known one to the web client. Viomi also (mostly?) pushes status updates to the connected dummycloud client so polling may technically not be necessary (except immediately after startup which was also disabled by your change).

PR #454, opened 2020-03-08
Viomi: Map & misc updates

rumpeltux:
Mostly fixing map parsing & updating viomi documentation.

Some general fixes around verbosity of logging (may need to adopt the `trace`-command set for Roborock if so desired).

### DIFF
--- a/lib/ViomiMapParser.js
+++ b/lib/ViomiMapParser.js
@@ -1,3 +1,4 @@
+const Logger = require("./Logger");
 
 /**
  * Experimental parser for the Viomi map binary format.
@@ -139,13 +140,13 @@ ViomiMapParser.prototype.parse = function() {
 
     if (historyMask & 0x10) {
         let head = asInts(this.take(12, "virtual wall"));
-        console.log("VirtualWall", head);
+        Logger.trace("VirtualWall", head);
         this.take(92 * head[2], "virtual wall content");
     }
 
     if (historyMask & 0x20) {
         let head = asInts(this.take(12, "area head"));
-        console.log("AreaClean", head);
+        Logger.trace("AreaClean", head);
         this.take(92 * head[2], "area content");
     }
 
@@ -184,7 +185,7 @@ ViomiMapParser.prototype.parse = function() {
 
 /** Debug log name and binary location. */
 ViomiMapParser.prototype.log = function(name, extra) {
-    // console.debug(name, (this.offset).toString(16), extra);
+    Logger.trace(name, (this.offset).toString(16), extra);
 };
 
 ViomiMapParser.prototype.parsePose = function() {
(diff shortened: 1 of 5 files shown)


lib/ViomiMapParser.js
```diff
@@ -183,7 +184,7 @@ ViomiMapParser.prototype.parse = function() {
 
 /** Debug log name and binary location. */
 ViomiMapParser.prototype.log = function(name, extra) {
-    console.debug(name, (this.offset).toString(16), extra);
+    // console.debug(name, (this.offset).toString(16), extra);
```
alexkn: Why is this commented out?
You should use the Logger class.

lib/ViomiMapParser.js
```diff
@@ -208,17 +209,42 @@ ViomiMapParser.prototype.parsePose = function() {
...
+    this.take(4, "room header");
+    let maps = {};
+    /*
+    Some examples:
+    e3 5a 18 5e 00 01 00 00 00
+    e3 5a 18 5e 00 00 00 00 00
+    e3 5a 18 5e 03[41 6c 74]b4 d2 63 5e 07[5a 75 68 61 75 73 65]01 00 00 00> // 'Alt', 'Zuhause'
+    e3 5a 18 5e 07[e5 9c b0 e5 9b be 31]01 00 00 00 '地图1'
+     */
+    // eslint-disable-next-line no-constant-condition
```
alexkn: Maybe use a do...while loop?

lib/devices/Viomi.js
```diff
@@ -102,7 +105,20 @@ class Viomi extends MiioVacuum {
...
 
     parseMap(data) {
-        return new ViomiMapParser(data).parse();
+        try {
+            return new ViomiMapParser(data).parse();
+        } catch (e) {
+            // save data for debugging
+            let i = 0;
+            let filename = "";
+            // eslint-disable-next-line no-constant-condition
```
alexkn: Maybe use a do...while loop?

lib/miio/MiioSocket.js
```diff
@@ -153,11 +161,21 @@ class MiioSocket {
...
             this.socket.send(packet, 0, packet.length, this.rinfo.port, this.rinfo.address);
         });
     }
 
+    /**
+     * @param {string} method
+     * @param {string} msg
+     * @param {any[]} args
+     */
+    traceOrDebug(method, msg, ...args) {
```
alexkn: Why is that method needed?
Add a comment

rumpeltux:
Thanks for the comments. Addressed them in 9357935

schannall:
> They are deleted on reboot. Not deleting them is somewhat intentional so that users can pull them and file bug reports.

Will this be a Problem regarding free space on the robot? There does not seem to be too much free space. I can't check right now as I'm not having too much luck building the binary for the robot.

rumpeltux:
> > They are deleted on reboot. Not deleting them is somewhat intentional so that users can pull them and file bug reports.
> 
> Will this be a Problem regarding free space on the robot? There does not seem to be too much free space. I can't check right now as I'm not having too much luck building the binary for the robot.

Right now, Valetudo doesn’t restart automatically for viomi, so manual intervention / a reboot is required anyways. Once it matures, we can change the logic.

PR #457, opened 2020-03-14
[Viomi] Support robot-provided zones.

rumpeltux:
For Viomi the Robot itself keeps track of rooms.
Exposing those with the zones interface allows the UI to start cleaning them manually.

They are otherwise filtered from the UI. This should thus be a no-op for Roborock, except that the `// Todo: rename areas to zones in config` comment was addressed and some more type-safety was added.

### DIFF
--- a/client/home.js
+++ b/client/home.js
@@ -18,6 +18,8 @@ var robotStateDetailsTime = document.getElementById("robot-state-details-time");
 var loadingBarHome = document.getElementById("loading-bar-home");
 
 var config = {};
+/** @type {Array<{id:number, name:string}>} */
+var zones = [];
 
 var BUTTONS = {
     "start": {button: startButton, url: "api/start_cleaning"},
@@ -136,19 +138,16 @@ function handleGoToButton() {
 // eslint-disable-next-line no-unused-vars
 function handleAreaButton() {
     window.clearTimeout(currentRefreshTimer);
-    var options = [];
-
-    for (var i = 0; i < config.areas.length; i++) {
-        options.push(config.areas[i][0]);
-    }
+    /** @type {Array<string|object>} */
+    var options = zones ? zones.map(z => z.name) : [];
 
     options.push({label: "Cancel", icon: "md-close"});
 
     ons.openActionSheet({title: "Clean area", cancelable: true, buttons: options})
         .then(function(index) {
-            if (index > -1 && index < config.areas.length) {
+            if (index > -1 && zones && index < zones.length) {
                 fn.requestWithPayload(
-                    "api/start_cleaning_zone_by_name", JSON.stringify([config.areas[index][0]]),
+                    "api/start_cleaning_zones_by_id", JSON.stringify([zones[index].id]),
                     "PUT", function(err) {
                         if (err) {
                             ons.notification.toast(err, {buttonLabel: "Dismiss", timeout: 1500});
@@ -245,9 +244,12 @@ ons.getScriptPage().onShow = function() {
         if (config.spots)
             if (config.spots.length > 0)
                 goToButton.removeAttribute("disabled");
-        if (config.areas)
-            if (config.areas.length > 0)
-                areaButton.removeAttribute("disabled");
+    });
+    fn.request("api/zones", "GET", (err, res) => {
+        if (res && res.length > 0) {
+            zones = res;
+            areaButton.removeAttribute("disabled");
+        }
     });
     updateHomePage();
 };
--- a/lib/Configuration.js
+++ b/lib/Configuration.js
@@ -22,20 +22,10 @@ class Configuration {
             Logger.info("Loading configuration file:", this.location);
 
             try {
-                this.settings = Object.assign(this.settings, JSON.parse(fs.readFileSync(this.location, {"encoding": "utf-8"})));
-                // Migrate legacy "areas" to "zones"
-                if (this.settings["zones"] === undefined && this.settings["areas"] !== undefined) {
-                    /** @type {Array<Configuration.Zone>} */
-                    const zones = this.settings["areas"].map(
-                        (v, i) => ({
-                            id: -(i + 1),
-                            name: v[0],
-                            areas: v[1] ? Array.isArray(v[1]) ? v[1] : [v[1]] : [],
-                            user: true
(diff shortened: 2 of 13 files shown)


rumpeltux:
Any more comments for this?

client/home.js
```diff
@@ -136,19 +138,16 @@ function handleGoToButton() {
...
-    }
+    /** @type {Array<string|object>} */
+    var options = zones ? zones.map(z => z.name) : [];
 
     options.push({label: "Cancel", icon: "md-close"});
 
     ons.openActionSheet({title: "Clean area", cancelable: true, buttons: options})
         .then(function(index) {
-            if (index > -1 && index < config.areas.length) {
+            if (index > -1 && zones && index < zones.length) {
```
schannall: is the && zones && needed? Wouldn't it need a parameter?

lib/Configuration.js
```diff
@@ -23,6 +23,19 @@ class Configuration {
 
             try {
                 this.settings = Object.assign(this.settings, JSON.parse(fs.readFileSync(this.location, {"encoding": "utf-8"})));
+                // Migrate legacy "areas" to "zones"
```
schannall: Should this be in the Configuration.js or would it be better to have it somewhere else? This should be a viomi function? As such I think it should be somewhere else. (Just my opinion, feel free to disagree and ignore)

lib/devices/Roborock.js
```diff
@@ -5,7 +5,7 @@ const Tools = require("../Tools");
 const RRMapParser = require("../RRMapParser");
 
 /**
- * @implements {MiioVacuum}
+ * @extends {MiioVacuum}
```
schannall: Same here. Maybe we should start early to have a nev device.js - the viomi is not a Roborock.
Maybe have a "device-independent", "Roborock", "MiioVacuum"?

lib/res/default_settings.json
```diff
@@ -1,6 +1,5 @@
 {
     "spots": [],
-    "areas": [],
```
alexkn: shouldn't there be an empty "zones" array now?

lib/webserver/WebServer.js
```diff
@@ -447,10 +447,10 @@ const WebServer = function (options) {
     });
 
     /* New function with support of current zone cleanup */
-    this.app.put("/api/start_cleaning_zone_by_name", async (req, res) => {
+    this.app.put("/api/start_cleaning_zones_by_id", async (req, res) => {
```
alexkn: why you changed the api from names to ids?
I think its better to stay with names.

lib/res/default_settings.json
rumpeltux: The Configuration.js code deals with undefined, so not necessary to have this entry.

lib/webserver/WebServer.js
rumpeltux: Both Roborock and Viomi APIs operate on IDs. The UI operates on IDs. If users were to give different zones the same name (or the robot-provided name matches a user-defined name) we run into problems. I think it’s much safer to go with IDs.

client/home.js
rumpeltux: We need to check that zones is a valid object before checking it’s length (e.g. in case it’s null).

lib/Configuration.js
rumpeltux: This is to switch users of older Valetudo versions seamlessly to the new format. Maybe worth to extract a separate method though.

lib/devices/Roborock.js
rumpeltux: MiioVacuum is device-independent.
Viomi and Roborock are its implementations. The goal is that both conform to the same interface (not yet 100% there).

lib/devices/Roborock.js
```diff
@@ -505,33 +505,40 @@ class Roborock extends MiioVacuum {
...
-            let area = areas.find(e => Array.isArray(e) && e[0] === zone_id);
-            if (area && Array.isArray(area[1])) {
-                area[1].forEach(function (zone) {
-                    zonesToClean.push(zone);
-                });
-                zoneNames.push(area[0]);
+        zone_ids.forEach((zone_id) => {
+            let zone = zones.get(zone_id);
+            if (zone) {
+                areasToClean.concat(zone.areas);
```
alexkn: Array.concat() returns the new array and not modifies the existing arrays.
areasToClean = areasToClean.concat(zone.areas);

lib/webserver/WebServer.js
```diff
@@ -461,23 +461,29 @@ const WebServer = function (options) {
...
-                zone.name,
-                zone.coordinates
-            ]);
-            this.configuration.set("areas", zones);
+            const zones = this.configuration.getZones();
+            const found = new Set();
+            req.body.map(zone => {
+                /** @type {Object} */
+                const oldZone = zones.get(zone.id);
+                if (oldZone) {
```
alexkn: This drops newly created zones.

rumpeltux: Yes, already fixed in recent commit. Thanks!

lib/devices/Roborock.js
rumpeltux: Thanks for catching!

lib/MqttClient.js
```diff
@@ -384,7 +384,7 @@ class MqttClient {
                  */
                 case CUSTOM_COMMANDS.ZONED_CLEANUP:
                     if (Array.isArray(msg.zone_ids) && msg.zone_ids.length) {
-                        this.vacuum.startCleaningZoneByName(msg.zone_ids)
+                        this.vacuum.startCleaningZonesById(msg.zone_ids)
```
alexkn: This broke my node-red interface.
MQTT should use the zone names.
I don't want to use the ids here.

rumpeltux: Fair. Done.

lib/Configuration.js
rumpeltux: Done.

leonardpitzu:
issue persists - zone editor is empty also with the last commit

error point to forbidden-markers-configuration-map.js:2
`SyntaxError: import declarations may only appear at top level of a module`

PR #460, opened 2020-03-19
add type checking with type script compiler

alexkn:
with initial fixes

replaced prettycron with cronstrue, cause there are no valid type definitions for prettycron and it seems not maintained

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -18,6 +18,7 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm install
+    - run: npm run ts-check
     - run: npm run lint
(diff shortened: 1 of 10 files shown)


rumpeltux:
Thanks! This generally LGTM.

I feel a little wary wrt/ the including the Roborock-specific interfaces there. Maybe we should mark them as "may-need-refactoring" or so. Can also do that later.

Just to confirm: the updates to `yarn.lock` are intentional right?

alexkn:
I don't see the abstract methods as "Roborock-specific".
They are currently expected by the web/mqtt api.
But yes, they should be refactored later with some system to handle the device specific functions. 

The yarn.lock get automatically updated when added the new dependencies.
I think that's correct.

PR #467, opened 2020-04-01
api clean up

alexkn:
TODO:
- [x] Roborock Implementation 
- [x] Viomi Implementation

### DIFF
--- a/lib/devices/MiioVacuum.js
+++ b/lib/devices/MiioVacuum.js
@@ -13,15 +13,14 @@ const NotImplementedError = require("./NotImplementedError");
  */
 class MiioVacuum {
     /**
-    * @constructor
-    * @param options {object}
-    * @param options.cloudKey {Buffer} The pre-shared unique key of your robot
-    * @param options.configuration {import('../Configuration')}
-    * @param options.deviceId {string} The unique Device-id of your robot
-    * @param options.events {import("events").EventEmitter}
-    * @param options.ip {string} Remote IP
-    * @param options.tokenProvider {()=>Buffer}
-    */
+     * @param {object} options
+     * @param {Buffer} options.cloudKey The pre-shared unique key of your robot
+     * @param {import('../Configuration')} options.configuration
+     * @param {string} options.deviceId The unique Device-id of your robot
+     * @param {import("events").EventEmitter} options.events
+     * @param {string} options.ip Remote IP
+     * @param {()=>Buffer} options.tokenProvider
+     */
     constructor(options) {
         this.status = new Status();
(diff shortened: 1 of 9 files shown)


lib/devices/MiioVacuum.js
```diff
@@ -70,49 +69,84 @@ class MiioVacuum {
...
+     * @returns {Promise<void>}
+     */
+    async pollStatus() {
+        throw new NotImplementedError();
+    }
 
     /**
      * Returns the current status of the robot
+     *
+     * @public
```
rumpeltux, line 104: isn’t public the default already?

lib/devices/MiioVacuum.js
```diff
@@ -309,18 +365,36 @@ class MiioVacuum {
     }
 
     /**
+     * @typedef AppLocale
+     * @property {string} name
+     * @property {string} bom
```
rumpeltux, line 370: what’s that?

lib/devices/MiioVacuum.js
```diff
@@ -309,18 +365,36 @@ class MiioVacuum {
...
+     * @returns {Promise<AppLocale>}
      */
     async getAppLocale() {
         throw new NotImplementedError();
     }
 
+    /**
+     * @typedef Timer
+     * @property {string} id
+     * @property {string} cron
```
rumpeltux, line 388: I assume that’s the timer-spec in crontab format?

lib/webserver/ApiRouter.js
```diff
@@ -27,6 +27,7 @@ const removeOldUploads = async (extension = null) => {
  * Api routes used by the webinterface
  *
  * @param {import("./WebServer")} webserver
+ * @returns {import("express").Router}
```
rumpeltux, line 30: Type is already imported so `@returns {express.Router}` should just work, right?

lib/devices/MiioVacuum.js
alexkn: Yes, public is the default. I simply added the access modifiers everywhere to make it clearer.

lib/devices/MiioVacuum.js
alexkn: I don't know.
Roborock returns 'bom': 'A.03.0005'

lib/devices/MiioVacuum.js
alexkn: Yes

lib/webserver/ApiRouter.js
alexkn: that's right

PR #477, opened 2020-04-11
RRMapParser for firmware 2008 / gen3 maps

alexkn:
with tests
and fix parsing no_go_areas


lib/RRMapParser.js
```diff
@@ -129,18 +158,42 @@ class RRMapParser {
...
-            },
-            dimensions: {
-                height: block.view.readInt32LE(0x10),
-                width: block.view.readInt32LE(0x14)
+    /**
+     * @param {Block} block
+     */
+    static PARSE_IMAGE_BLOCK(block) {
+        const parsedBlock = {
+            segments: { //TODO: naming
```
Hypfer, line 166: Is "segments" the right term for this kind of data?

PR #478, opened 2020-04-13
move fan speed presets to vacuum implementation

alexkn:
and support for new firmware

### DIFF
--- a/lib/devices/Roborock.js
+++ b/lib/devices/Roborock.js
@@ -11,6 +11,11 @@ class Roborock extends MiioVacuum {
     constructor(options) {
         super(options);
         this.lastMapPoll = new Date(0);
+        /**
+         * @private
+         * @type {{[id: string]: {label: string, value: number}}}
+         */
+        this.fanSpeeds = null;
     }
 
     onMessage(msg) {
@@ -115,6 +120,10 @@ class Roborock extends MiioVacuum {
 
 
     updateStatus(res) {
+        if (!this.fanSpeeds) {
+            this.detectFanSpeeds(res["msg_ver"]);
+        }
+
         this.status.state = STATUS_MAP[res["state"]];
         this.status.battery = res["battery"];
         this.status.clean_time = res["clean_time"];
@@ -122,7 +131,7 @@ class Roborock extends MiioVacuum {
         this.status.error_code = res["error_code"];
         this.status.map_present = res["map_present"];
         this.status.in_cleaning = res["in_cleaning"];
-        this.status.fan_power = res["fan_power"];
+        this.status.fan_power = Object.keys(this.fanSpeeds).find(key => this.fanSpeeds[key].value === res["fan_power"]);
         this.status.dnd_enabled = res["dnd_enabled"];
         this.status.lab_status = res["lab_status"];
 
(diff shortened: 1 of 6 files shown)


lib/devices/Roborock.js
```diff
@@ -335,18 +344,67 @@ class Roborock extends MiioVacuum {
...
+            "mop": {label: "Mop", value: 105},
+        };
+
+        if (msg_ver >= 3) {
+            delete fanSpeeds.whisper;
+            fanSpeeds.min.value = 101;
+            fanSpeeds.medium.value = 102;
+            fanSpeeds.high.value = 103;
+            fanSpeeds.max.value = 104;
+            fanSpeeds.mop.value = 105;
```
rumpeltux, line 366: Should use quotes consistently, e.g.

    delete fanSpeeds["whisper"];
    fanSpeeds["min"].value = …

lib/devices/Roborock.js
```diff
@@ -335,18 +344,67 @@ class Roborock extends MiioVacuum {
...
+     * @private
+     */
+    detectFanSpeeds(msg_ver) {
+        let fanSpeeds = {
+            "whisper": {label: "Whisper", value: 1},
+            "min": {label: "Quiet", value: 38},
+            "medium": {label: "Balanced", value: 60},
+            "high": {label: "Turbo", value: 75},
+            "max": {label: "Max", value: 100},
+            "mop": {label: "Mop", value: 105},
```
rumpeltux, line 357: Are you sure you want to model mopping as a "fan-speed"?

lib/devices/Roborock.js
```diff
@@ -115,14 +120,18 @@ class Roborock extends MiioVacuum {
...
+
         this.status.state = STATUS_MAP[res["state"]];
         this.status.battery = res["battery"];
         this.status.clean_time = res["clean_time"];
         this.status.clean_area = res["clean_area"];
         this.status.error_code = res["error_code"];
         this.status.map_present = res["map_present"];
         this.status.in_cleaning = res["in_cleaning"];
-        this.status.fan_power = res["fan_power"];
+        this.status.fan_power = Object.keys(this.fanSpeeds).find(key => this.fanSpeeds[key].value === res["fan_power"]);
```
rumpeltux, line 134: I’d argue, `Status.fan_power` should map to an arbitrary percentage value (e.g. 0-100) rather than the device-specific preset?
The UI would gather the presets, but otherwise the API purposes we’d just deal with the percentage value?
While at it: should also document `Status.fan_power`.

lib/devices/Roborock.js
Hypfer: This sadly is how roborock did it :/

rumpeltux: I understand, but that’s something I think the Valetudo API could abstract away right?
Roborock.js can know about this, but in the UI the mopping option should probably not be available via the fan-speed button?
* For Viomi I added support to enable mopping automatically based on installed box (#474). Not sure if something like that would be feasible for RR as well?
* For Viomi there are two settings, mixed (mop&vacuum) or mop only, which is currently not exposed through the UI, so a mopping-settings UI would still make sense.

(We can file this a feature request and leave this out of this refactoring for now).

lib/devices/Roborock.js
Hypfer: +1 to that

alexkn: It looks as if the 2008 firmware no longer supports the percent values.
Then I have to map the presets to the old values.

lib/devices/Roborock.js
alexkn: fixed

lib/devices/Roborock.js
alexkn: The web interface only uses the presets and mqtt also only used the names before.
I think we should stay with names instead of percentage values.

lib/devices/Roborock.js
alexkn: I think you should propose the webinterface/API changes for viomi and then we look to adapt the roborock implementation.

lib/devices/Roborock.js
Hypfer: I think it would be okay to have a fixed set of valetudo vacuum fan_power constants and just map the device specific ones to them.

Like Min, Silent, Normal, Max and Turbo for example. If the device supports less, we only provide some of those. If it supports more we ignore the additional options.

alexkn: done

PR #480, opened 2020-04-13
MQTT go_to command

figurcoe:
added possibility to give the spot_id as coordinates in addition to the name.

Something similar could be done for zoned cleanup to be fully compatible with
https://github.com/PiotrMachowski/Lovelace-Xiaomi-Vacuum-Map-card

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -394,6 +394,16 @@ class MqttClient {
                 }
                 /**
                  * {
+                 *   "command": "app_segment_clean",
+                 *   "params": [room_id]
+                 * }
+                 */
+                case CUSTOM_COMMANDS.APP_SEGMENT_CLEAN: {
+                    Logger.info("Custom command 'app_segment_clean' is currently not supported");
+                    break;
+                }
+                /**
+                 * {
                  *   "command": "app_zoned_clean",
                  *   "params": [[x1, y1, x2, y2, iterations],..]
                  * }
@@ -437,11 +447,6 @@ class MqttClient {
                  * }
                  */
                 case CUSTOM_COMMANDS.GO_TO:
-                    // TODO remove
-                    // msg.spot_id is an array of exactly 2 values considered to be coordinates
-                    // if (Array.isArray(msg.spot_id) && msg.spot_id.length === 2) {
-                    //     this.vacuum.goTo(msg.spot_id[0], msg.spot_id[1]);
-                    // } else
                     if (msg.spot_id) {
                         const spots = this.configuration.get("spots");
                         const spot_coords = spots.find(e => Array.isArray(e) && e[0] === msg.spot_id);
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -383,12 +383,7 @@ class MqttClient {
                  *   "params": [x, y]
                  * }
                  */
-                // entity_id: vacuum.rose       <---------------------------------------------------------------------------------------
-                // command: app_goto_target     <---------------------------------------------------------------------------------------
-                // params:                      <---------------------------------------------------------------------------------------
-                //   params: [11,22]            <---------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


lib/MqttClient.js
```diff
@@ -404,9 +404,19 @@ class MqttClient {
...
+                 * or (with xxx and yyy being coordinates)
+                 *
+                 * {
+                 *   "command": "go_to",
+                 *   "spot_id": [xxx, yyy]
+                 * }
                  */
                 case CUSTOM_COMMANDS.GO_TO:
-                    if (msg.spot_id) {
+                    // msg.spot_id is an array of exactly 2 values considered to be coordinates
```
alexkn, line 416: Try to rework the existing code, so there is only one this.vacuum.goTo call in there.

figurcoe:
Hi,
Didn't want to change too much. I'll see what I can do.
Br

Am April 13, 2020 8:04:22 PM UTC schrieb "Alexander Knöbel" <notifications@github.com>:

figurcoe:
I'll give it a look and reopen it when potentially finished.

figurcoe:
Sorry for the inconvenience, I'm currently not able to test/check this. But as it relies on internal functions, no big problems should occure. I added "new" custom commands "app_goto_target" and "app_zoned_clean".

lib/MqttClient.js
```diff
@@ -13,8 +13,11 @@ const MQTT_COMMANDS = {
 };
 
 const CUSTOM_COMMANDS = {
+    APP_GOTO_TARGET: "app_goto_target",
+    APP_SEGMENT_CLEAN: "app_segment_clean",
```
alexkn, line 17: APP_SEGMENT_CLEAN is not used in your code.
Valetudo currently does not support segment cleaning.

lib/MqttClient.js
```diff
@@ -406,6 +437,11 @@ class MqttClient {
                  * }
                  */
                 case CUSTOM_COMMANDS.GO_TO:
+                    // TODO remove
```
alexkn, line 440: remove unused code

Hypfer:
I think that https://github.com/TheLastProject/lovelace-valetudo-map-card would be better-suited for this job, since it does actually receive the map as data and therefore doesn't need `calibration_points` like the `lovelace-xiaomi-vacuum-map-card`.

I did leave out this feature, because at the time there was no reasonable way to know the coordinates anyways, but with the lovelace valetudo card it would definitely make sense _I think_

What is your opinion on this @TheLastProject ?

lib/MqttClient.js
figurcoe: Will do, somehow this slipped through.

lib/MqttClient.js
figurcoe: Yeah, this obviously true, my bad. I realized that, then unfortunately forgot about this.

How about keeping it in there, but adding a case logging a message saying it is not supported yet. This way it would be indicated that it is part of the api, generally supported by other software, but not implemented with Valetudo (maybe yet).

figurcoe:
I would definitely like to see the features of [Lovelace-Xiaomi-Vacuum-Map-card](https://github.com/PiotrMachowski/Lovelace-Xiaomi-Vacuum-Map-card) implemented in [lovelace-valetudo-map-card](https://github.com/TheLastProject/lovelace-valetudo-map-card). I would offer to help, but as you can see, I don't have any clue about javascript and I'm pretty busy. I fear, I'd cause way more trouble than I'd actually help.

Anyways, I believe these features would be a worthy addition.

Thanks a lot, for what is already available, as of today!

TheLastProject:
I'm not completely sure what you're asking me @Hypfer. A way to define spots in my Home Assistant card qnd sending a call with coordinates from my card?

I don't know if or when I'd get to implementing that as JavaScript isn't my language of choice and I have many other projects, but I'd definitely welcome pull requests.

figurcoe:
So, just for my understanding, what shall I change?

PR #485, opened 2020-04-16
fix addTimer for firmware 2008

alexkn:


lib/devices/Roborock.js
```diff
@@ -271,7 +274,11 @@ class Roborock extends MiioVacuum {
      * @param {string} cron
      */
     async addTimer(cron) {
-        await this.sendLocal("set_timer", [[Date.now().toString(),[cron, ["",""]]]], {});
+        if (this.isGen3) {
+            await this.sendLocal("set_timer", [[Date.now().toString(), [cron, ["start_clean", {"fan_power": 102, "segments": "", "repeat": 1, "clean_order_mode": 1}]]]], {});
```
rumpeltux, line 278: (nit) add some wrap to prevent too long lines

rumpeltux, line 278: Shouldn’t `fan_power` reference the current fan setting?

lib/devices/Roborock.js
```diff
@@ -16,6 +16,8 @@ class Roborock extends MiioVacuum {
          * @type {{[id: string]: {label: string, value: number}}}
          */
         this.fanSpeeds = null;
+        /** @private */
```
rumpeltux, line 19: Add some comments on what this means / is being used for.
And potentially when using the setting latter, add some comments about the differences in command set (e.g. pre-gen3 don’t take arguments for set_timer?).

lib/devices/Roborock.js
alexkn: I do not think so. 
You should actually be able to set it when creating the timer in the web interface, but this PR was intended as a bug fix only.

lib/devices/Roborock.js
alexkn: done

lib/devices/Roborock.js
alexkn: done

PR #486, opened 2020-04-17
fix maps with segments

alexkn:


lib/RRMapParser.js
```diff
@@ -216,16 +216,19 @@ class RRMapParser {
                         parsedBlock.dimensions.height-1 - Math.floor(i / parsedBlock.dimensions.width)
                     ];
 
-                    switch (val) {
+                    let type = val & 0b00000111;
```
Hypfer, line 219: This is an awesome idea 👍

PR #487, opened 2020-04-18
Fix consecutive page loading

alexkn:


client/settings-persistent-data.js
```diff
@@ -2,7 +2,7 @@
 
 var loadingBarSettingsPersistentData =
     document.getElementById("loading-bar-settings-persistent-data");
-const labMode = /** @type {HTMLInputElement} */ (document.getElementById("lab_mode_enabled"));
+var labMode = /** @type {HTMLInputElement} */ (document.getElementById("lab_mode_enabled"));
```
rumpeltux, line 5: let?

alexkn: I tried, but let also throws a redeclaration error there.

PR #490, opened 2020-04-26
Docs navigation

alexkn:


Hypfer:
I think that it would make sense to have a seperate subfolder for installation guides since there will be most likely more of that later

alexkn:
done (I thought it was not that easy)

PR #496, opened 2020-04-29
Support changing MQTT settings in web UI

TheLastProject:
Screenshot_2020-04-29 Mi Robot Vacuum
Screenshot_2020-04-29 Mi Robot Vacuum(1)

Must say, the dev env is actually pretty nice :)

### DIFF
--- a/client/settings-mqtt.html
+++ b/client/settings-mqtt.html
@@ -1,0 +1,126 @@
+<ons-page id="settings-mqtt-page">
+    <ons-toolbar>
+        <div class="left">
+            <ons-back-button>Settings</ons-back-button>
+        </div>
+        <div class="center">MQTT</div>
+        <div class="right">
+        </div>
+    </ons-toolbar>
+    <ons-progress-bar id="loading-bar-settings-mqtt" value="0" indeterminate="indeterminate"></ons-progress-bar>
+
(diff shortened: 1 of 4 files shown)


Hypfer:
There should probably also be a way to apply that configuration.
As of now, enabling will only be in effect on the next reboot

We should at least have a hint about that because that will surely confuse users

TheLastProject:
Oops, I completely forgot about that part...

TheLastProject:
In the Telegram chat, it was discussed how to best apply the MQTT changes on runtime. Restarting only the MQTT part was considered the best solution, but this is not currently supported by the code.

Leaving this note here as it's a blocker for this to be merged.

Hypfer:
Since enabling control of the MQTT Interface via the Webinterface is a larger change, I'd suggest that we just add a hint to the UI regarding the necessary reboot.

This is already a huge improvement. Especially since rebooting the robot doesn't necessarily require ssh access.

TheLastProject:
How would you reboot the vacuum without SSH? Just hold the power button for several seconds?

Hypfer:
Exactly. If the english voicepack it will respond with `Turning off` and turn off the LED after a few seconds.
Then you can simply turn it on again

TheLastProject:
Done :)

PR #502, opened 2020-05-02
fix: gen1 fanspeeds

### LINKED ISSUE #499: Power mode is undefined after 0.5.0 update
### Screenshots
Screenshot_20200430_141111

### Vacuum Model
Gen1

### Valetudo Version
0.5.0

alexkn:
should fix #499


Hypfer:
We should probably think about letting each roborock vacuum have their own class in the long run

Hypfer:
What about

https://github.com/Hypfer/Valetudo/blob/3ca66202c2bc859c015d2f0f84bb3340e4eb06fe/lib/devices/Roborock.js#L132

?

PR #507, opened 2020-05-05
Expose zone changes over Mqtt

### LINKED ISSUE #102: Coordinates of the zones in map
would be nice to see the coordinates of the zone in the map

TheLastProject:
Fixes #102.

Broadcasts to `/zones` on change, same data as `/api/zones` but prefixed with "zones" so that Home Assistant is okay with parsing it as a JSON attribute:
image

Example Home Assistant configuration.yaml:
```yaml
sensor:
  - platform: mqtt
    state_topic: "valetudo/rockrobo/state"
    json_attributes_topic: "valetudo/rockrobo/zones"
    name: xiaomi_map_zones
    value_template: 'OK'
    scan_interval: 5
```

image

(P.S.: What's the 1 as the last value of an area and why is id null and -1?)

### DIFF
--- a/lib/Events.js
+++ b/lib/Events.js
@@ -17,6 +17,17 @@ class Events {
         this.eventEmitter.on(Events.MapUpdated, listener);
     }
 
+    emitZonesChanged() {
+        this.eventEmitter.emit(Events.ZonesChanged);
+    }
+
+    /**
+     * @param {() => void} listener
+     */
+    onZonesChanged(listener) {
+        this.eventEmitter.on(Events.ZonesChanged, listener);
+    }
+
     /**
      * @param {import("./miio/Status")} status
      */
@@ -44,6 +55,7 @@ class Events {
 }
 
 Events.MapUpdated = "MapUpdated";
+Events.ZonesChanged = "ZonesChanged";
 Events.StatusUpdated = "StatusUpdated";
 Events.MqttConfigChanged = "MqttConfigChanged";
 
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -80,6 +80,12 @@ class MqttClient {
             }
         });
 
+        this.events.onZonesChanged(() => {
+            if (this.provideMapData === true) {
+                this.updateZonesTopic();
+            }
+        });
+
         this.events.onStatusUpdated((statusData) => {
             this.updateStatusTopic(statusData);
             this.updateAttributesTopicOnEvent(statusData);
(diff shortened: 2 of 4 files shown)


Hypfer:
https://github.com/Hypfer/Valetudo/blob/9efa812982199603ea851cb9a651dc0a9f9e1600/lib/devices/Roborock.js#L561

The 1 as the last value is the number of times the zone should be cleaned. That's currently missing in the UI => #521

https://github.com/Hypfer/Valetudo/blob/4f7ff43bf3ef211f5bf0ea6f41be369bd454c117/lib/Configuration.js#L87

The -1 is there because the zone was migrated from the old format. I guess it's negative to prevent collisions with newly created ones? @rumpeltux

rumpeltux:
> The -1 is there because the zone was migrated from the old format. I guess it's negative to prevent collisions with newly created ones? @rumpeltux

See this comment: https://github.com/Hypfer/Valetudo/blob/4f7ff43bf3ef211f5bf0ea6f41be369bd454c117/lib/Configuration.js#L107

For Viomi there are also zones stored by the native Robot App, these have positive IDs.

rumpeltux:
More generally I would propose that instead of sending the Valetudo internal representation via MQTT that we perform some preprocessing so that we only return the data that remote needs and can update the representation format.e.g. I think it would be quite bad if HA were to parse the areas array in which the last value is the repeat count, instead we should expose an explicit `"repeats":` field or so.

TheLastProject:
Okay, a repeats field, sounds good. Any other preprocessing you'd like me to implement? I'll implement it for both the API and MQTT to keep them consistent.

rumpeltux:
> Okay, a repeats field, sounds good. Any other preprocessing you'd like me to implement? I'll implement it for both the API and MQTT to keep them consistent.

Depends on what HA will use this for. The mm relative to [10m,10m] also seems not to be any kind of standard unit, so you probably would want to convert this as well. Maybe areas should consist of polygons instead of rects?

And you probably don’t need to pass the `user` attribute.

TheLastProject:
It's indeed not a standard unit, but it's the same type of value path.points, charger, robot and no_go_areas (as exposed through the API and MQTT) currently use, so it seems reasonable to stick to that. Or this is a time to normalize those values first before exposing a new attribute (zones) using these non-standard units too.

rumpeltux:
> It's indeed not a standard unit, but it's the same type of value path.points, charger, robot and no_go_areas (as exposed through the API and MQTT) currently use, so it seems reasonable to stick to that. Or this is a time to normalize those values first before exposing a new attribute (zones) using these non-standard units too.

I would argue that for new such APIs we should define a rough contract that future versions of Valetudo will try to keep (on a best effort basis).
The alternative would be keeping the status quo (exposing internals) with the risk of this breaking anytime when we do make changes to internal representation.

@Hypfer may also have an opinion on that though.

(And sorry for the long turnaround times in my responses…)

Hypfer:
I think that we should completely rework the MQTT Interface and then define said contract.
The attributes topic is a complete mess in the current version.

This would of course break existing applications/automations, but that would only happen once.

Hypfer:
Superseded by 76a9c5234dfbcce864d70d1d6db781af86d6a98b

PR #511, opened 2020-05-07
cleanup vacuum return types

alexkn:


lib/MqttClient.js
```diff
@@ -49,6 +49,15 @@ const HA_STATE_MAPPINGS = {
     "PAUSED": HA_STATES.PAUSED,
 };
 
+/**
+ * @param {number} number
+ * @param {number} precision
```
rumpeltux, line 54: Can you document what precision means? (e.g. `number of decimals after dot`)

PR #515, opened 2020-05-10
Generalize MiioSockets to support both server and client versions

rumpeltux:
This is in preparation of an experimental feature in which Valetudo proxies whitelisted requests/responses to the real xiaomi cloud which is great for development and debugging.

### DIFF
--- a/lib/Valetudo.js
+++ b/lib/Valetudo.js
@@ -138,16 +138,14 @@ class Valetudo {
         );
     }
 
+    /** @returns {number} */
     static DEVICE_ID_PROVIDER() { //TODO: merge with CLOUD_KEY_PROVIDER
         const did = this.readDeviceConf()["did"];
-
-        // noinspection JSConstructorReturnsPrimitive
-        return did ? did: "00000000"; //This doesnt work but it wont crash the system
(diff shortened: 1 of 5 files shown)


Hypfer:
This is not a good idea since it will bring more people to this project that certainly don't belong here at all. :/

It's Valetudo _or_ the cloud. We're not doing Valetudo + Cloud. I don't want these people here

rumpeltux:
I understand your concerns to some degree, but that said for me this is a very valuable development tool that I wouldn’t want to miss. Maybe as a middle ground such support could live in another branch and not be enabled by default.

In any case the changes in this PR are general in nature, so we also don’t need to finish the discussion now.

Hypfer:
I do agree that this is a helpful development tool but it also feels like pandora's box to me.

I'm certain that we'd see a fork which makes that happen and which also would further damage the name/brand/whatever you might call it Valetudo :/

Hypfer:
I guess there's no point in refraining from merging a useful feature just because it could also be used in _bad_ ways.

Valetudos mission statement is very clear about the no cloud part so that should hopefully be sufficient.

PR #517, opened 2020-05-11
Adds time and area information about current clean to MQTT

docBliny:
I wasn't able to find the current clean time nor cleaning area in the MQTT output, so I added these following the format in the Xiaomi Home Assistant integration (i.e. `cleaning_time` and `cleaned_area` from https://www.home-assistant.io/integrations/vacuum.xiaomi_miio/).

Allows showing active cleaning information in HA, such as the following card:
image

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -77,6 +77,8 @@ class MqttClient {
         this.last_ha_state = HA_STATES.IDLE;
         this.last_state = "UNKNOWN";
         this.last_attributes = {};
+        this.last_cleaned_area = 0;
+        this.last_cleaning_time = 0;
 
         this.loadConfig();
         if (this.enabled) {
@@ -240,6 +242,8 @@ class MqttClient {
     updateAttributesTopicOnEvent(statusData) {
         this.last_ha_state = HA_STATE_MAPPINGS[statusData.state];
         this.last_state = statusData.state;
+        this.last_cleaned_area = statusData.clean_area ? (statusData.clean_area / 1000000).toFixed(1) : 0;
+        this.last_cleaning_time = statusData.clean_time ? (statusData.clean_time / 60).toFixed(1) : 0;
 
         this.updateAttributesTopic();
     }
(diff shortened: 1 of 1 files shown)


lib/MqttClient.js
```diff
@@ -240,6 +242,8 @@ class MqttClient {
     updateAttributesTopicOnEvent(statusData) {
         this.last_ha_state = HA_STATE_MAPPINGS[statusData.state];
         this.last_state = statusData.state;
+        this.last_cleaned_area = statusData.clean_area ? (statusData.clean_area / 1000000).toFixed(1) : 0;
```
jonathan-reisdorf, line 245:
```suggestion
        this.last_cleaned_area = statusData.clean_area ? parseFloat((statusData.clean_area / 1000000).toFixed(1)) : 0;
```
otherwise the data type is sometimes string, sometimes number

jonathan-reisdorf:
Great PR, I was looking for this data in MQTT already as well. WDYT @Hypfer ?

lib/MqttClient.js
jonathan-reisdorf: Also we might increase the precision and leave it to the user consuming the MQTT to format accordingly

lib/MqttClient.js
```diff
@@ -266,6 +270,8 @@ class MqttClient {
                 response.cleanTime = round(cleanSummary.cleanTime, 1);
                 response.cleanArea = round(cleanSummary.cleanArea, 1);
                 response.cleanCount = cleanSummary.cleanCount;
+                response.cleaned_area = this.last_cleaned_area;
```
jonathan-reisdorf, line 273: the other attributes are camelCase here, better to be consistent

lib/MqttClient.js
Hypfer: Since these are rather imprecise floats, I don't think that that would make much sense to transmit data that will be certainly discarded anyways.

It's just ugly to look at random floating point imprecisions imho. Especially since the robot itself is around 0.1m² in size

Maybe .toFixed(2)? That would result in 10cm squares being the minimum resolution

docBliny:
Thanks all!

PR #523, opened 2020-05-17
refactor(ui): create api service

alexkn:
My long term plan is to rewrite the ui with Vue, but this cannot be done in one PR.
So I started with some refactoring:
- created a api service for all api requests
- use requests with async/await
- switching from XMLHttpRequest to fetch
- fixes for problems caused by using modules


Hypfer:
lgtm

PR #524, opened 2020-05-19
Update the reset map FAQ to include info for Gen1/Gen2 devices

axilleas:
I was following the instructions to enable persistent maps in the FAQ, but the map kept resetting after a full cleanup. I'm using a gen1 device and the info in the FAQ didn't apply to it. I later found out this comment https://github.com/dgiese/dustcloud/issues/211#issuecomment-491733796, which cleared my confusion.

This PR mentions the state for both gen1 and gen2 devices.

### DIFF
--- a/docs/_pages/faq.md
+++ b/docs/_pages/faq.md
@@ -96,8 +96,20 @@ server {
 ```
 
 ## How to keep map data after reboot
-In default the robot will lose map data after each reboot. To prevent the map reset use the mirobo command line tool (same tool you use for the installation process) and execute `mirobo raw-command set_lab_status 1`.
-This can be also done via filesystem access `echo -n "1" > /mnt/data/rockrobo/lab.cfg` on the robot itself.
+
+By default, the robot will lose map data after each reboot. For Gen2 devices,
+to prevent the map reset, use the `mirobo` command line tool (same tool you use
+for the installation process):
+
+```shell
+mirobo raw-command set_lab_status 1
+```
+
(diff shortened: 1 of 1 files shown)


docs/_pages/faq.md
```diff
@@ -96,23 +96,42 @@ server {
 ```
 
 ## How to keep map data after reboot
-In default the robot will lose map data after each reboot. To prevent the map reset use the mirobo command line tool (same tool you use for the installation process) and execute `mirobo raw-command set_lab_status 1`.
-This can be also done via filesystem access `echo -n "1" > /mnt/data/rockrobo/lab.cfg` on the robot itself.
+
+By default, the robot will lose map data after each reboot. For Gen2 devices,
```
axilleas, line 100: I'm not sure how this should be accomplished for gen1 devices :thinking: 

Is backing up and restoring `/mnt/data/rockrobo/last_map` enough? I haven't tested this to be sure.

Hypfer: Afaik it's not possible at all

axilleas: That's unfortunate :/ I'll resolve this then.

docs/_pages/faq.md
```diff
@@ -96,23 +96,42 @@ server {
...
+to prevent the map reset, use the `mirobo` command line tool (same tool you use
```
Hypfer, line 101: This is also integrated in valetudo since quite a while

Settings >Persistent Data

image

PR #526, opened 2020-05-27
Add Availability topic to let Home Assistant also detect an offline robot

Poeschl:
This MR adds a Last Will and Testament mechanism which tells Home Assistant about the current connection status.

The `online` status is only sent once at startup and the MQTT broker will set it to `offline` when the connection to Valetudo was lost.

### DIFF
--- a/lib/MqttClient.js
+++ b/lib/MqttClient.js
@@ -50,8 +50,8 @@ const HA_STATE_MAPPINGS = {
     "PAUSED": HA_STATES.PAUSED,
 };
 
-const HA_ONLINE_STATUS = 'online'
-const HA_OFFLINE_STATUS = 'offline'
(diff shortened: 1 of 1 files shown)


Hypfer:
image


Sadly, there seems to be something wrong with the feature since the daily reboots do seem to break availablity

Poeschl:
For me it's working fine wit the availability 😅 only a few Minutes "offline" at 3 am

Hypfer:
It should also publish it on each reconnect (if any) so i have no idea why this might be happening.

d96eb15 should be a mitigation

Poeschl:
Will look into it this evening 😉
Sending it on reconnect shouldn't be a big thing

Poeschl:
Which broker are you using?

Hypfer:
Mosquitto

Issue #382, opened 2019-11-27
Disable "Configure forbidden zones" on gen1


schlamar:
As this is not supported on gen1 devices, you should disable this option instead of returning a 501 error after the forbidden zone is saved (see #346 for details).

Hypfer:
Sounds reasonable. I'll await your PR

PR #534, opened 2020-06-07
Don't open forbidden zones editor if model is gen 1 

ne0phyte:
Fix for https://github.com/Hypfer/Valetudo/issues/382

When the "Configure forbidden zones" button is pressed on a gen 1, a toast is shown ("Not supported by your device").

I hope it's okay that the check is done when the button is pressed. 
Maybe it'd be better to disable/hide the button right away?

### DIFF
--- a/client/zones.html
+++ b/client/zones.html
@@ -10,14 +10,16 @@
     <ons-list id="spot-list">
         <ons-list-item>No spots are configured yet.</ons-list-item>
     </ons-list>
-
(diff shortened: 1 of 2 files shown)


Hypfer:
it would definitely be better to hide it in the first place

ne0phyte:
Updated to hide the item if model is rockrobo.vacuum.v1

Hypfer:
Now there's a race condition where you can click it if the API responds slow enough

ne0phyte:
Fixed by inverting condition and hiding by default.

PR #544, opened 2020-06-22
Core Data Structure Rework

Hypfer:
In the last few days, I tried to rework the whole internal state and map handling.

This is still WIP.

## Still To-Do
- [x] Update Roborock Methods to support new data format (e.g. zoned-cleanup, go-to, etc)
- [x] Fully update map rendering for new Format
- [x] Mop-related state + fan speeds
- [x] Re-introduce UI Button State
- [x] Port Viomi to new structure
- [x] Update Tests
- [ ] Publish final specs for other projects (e.g. lovelace card)
- [x] Various ToDos in the new Code

Maybe:

- [x] Update ICantBelieveItsNotValetudo
- [x] Write Docs on new core Concepts
- [ ] Create status diff or rework diff-usage
- [ ] Use the same entity classes in the UI
- [ ] Fully update client map to not assume roborock-specific coordinates/constants
- [x] Fix approximated robot angle
- [ ] Move current cleanup statistics somewhere else

## New Concepts
There's a new `SerializableEntities` class which shall be used for data that might be serialized.
By using it, one can simply call JSON.stringify on an instance and receive something that could be deserialized later.

Example: 
```
  {
    "__class": "BatteryAttribute",
    "metaData": {},
    "level": 100
  }
```

### Map
There's the ValetudoMap class now. See Comment + JSDoc:
https://github.com/Hypfer/Valetudo/blob/4eab2d3a782194ff689420e71ab7a27849368370/lib/entities/map/ValetudoMap.js#L3-L22

As hinted with the TODO, it should also contain logic such as calucation which layer intersects with which points etc.

#### MapLayer
See Comment + JSDoc:
https://github.com/Hypfer/Valetudo/blob/4eab2d3a782194ff689420e71ab7a27849368370/lib/entities/map/MapLayer.js#L3-L23

#### MapEntity
See Comment + JSDoc:
https://github.com/Hypfer/Valetudo/blob/4eab2d3a782194ff689420e71ab7a27849368370/lib/entities/map/MapEntity.js#L3-L15

### State
For the Robot's state there's the new RobotState class which has a reference to the latest map including all its map-related state information as well as an array of Attributes.

#### Attribute
Attributes provide a generic representation of possible robot state attributes.
Instead of having fixed field-names in the state schema, it is possible to simply push new nodes to the attributes array.

This allows different feature levels such as for example Mop data which isn't required for non-mop robots. Having some kind of model-based feature-flag system would work fine with this approach and might be the next valetudo extension.

##### StatusAttribute
Regarding attributes, the most relevant one is probably the StatusAttribute which is a generic approach to robot status information.

Inspired by the MQTT interface, these main states should include everything that might come.
https://github.com/Hypfer/Valetudo/blob/4eab2d3a782194ff689420e71ab7a27849368370/lib/entities/state/attributes/StatusAttribute.js#L26-L40

Furthermore, there are flags, which may contain additional information extending the Status Value
https://github.com/Hypfer/Valetudo/blob/4eab2d3a782194ff689420e71ab7a27849368370/lib/entities/state/attributes/StatusAttribute.js#L42-L54

For example there's the `RESUMABLE` flag which indicates that a `PAUSED` state could be unpaused.
Also, different cleaning scenarios don't have to be different states now. Instead there are flags for spot, zoned and section cleaning.

The class also contains logic to determine whether or not we should fast-poll the robots state
https://github.com/Hypfer/Valetudo/blob/4eab2d3a782194ff689420e71ab7a27849368370/lib/entities/state/attributes/StatusAttribute.js#L17-L22

I'd like to encourage everyone to take a look at the current state of this PR and provide feedback on the matter.

### Misc

By using a single array instead of an array of arrays for the roborock map parsing, memory consumption was cut in half in my tests.


lib/MqttClient.js
```diff
@@ -29,26 +29,19 @@ const HA_STATES = {
...
+
+const HA_STATE_MAPPINGS = {};
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.ERROR] = HA_STATES.ERROR;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.DOCKED] = HA_STATES.DOCKED;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.IDLE] = HA_STATES.IDLE;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.RETURNING] = HA_STATES.RETURNING;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.CLEANING] = HA_STATES.CLEANING;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.PAUSED] = HA_STATES.PAUSED;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.MANUAL_CONTROL] = HA_STATES.CLEANING;
+HA_STATE_MAPPINGS[attributes.StatusAttribute.VALUE.MOVING] = HA_STATES.CLEANING;
```
DeltaEvo, line 44:
```suggestion
const HA_STATE_MAPPINGS = {
    [attributes.StatusAttribute.VALUE.ERROR]: HA_STATES.ERROR,
    [attributes.StatusAttribute.VALUE.DOCKED]: HA_STATES.DOCKED,
    [attributes.StatusAttribute.VALUE.IDLE]: HA_STATES.IDLE,
    [attributes.StatusAttribute.VALUE.RETURNING]: HA_STATES.RETURNING,
    [attributes.StatusAttribute.VALUE.CLEANING]: HA_STATES.CLEANING,
    [attributes.StatusAttribute.VALUE.PAUSED]: HA_STATES.PAUSED,
    [attributes.StatusAttribute.VALUE.MANUAL_CONTROL]: HA_STATES.CLEANING,
    [attributes.StatusAttribute.VALUE.MOVING]: HA_STATES.CLEANING
};
```

lib/devices/MiioVacuum.js
```diff
@@ -19,10 +19,13 @@ class MiioVacuum {
      * @param {import("../miio/Model")} options.model
      */
     constructor(options) {
-        this.status = new Status();
+        this.robotState = new entities.state.RobotState({
+            map: require("../res/default_map")
```
DeltaEvo, line 23: What about removing the "default_map" from the server and add a way for the server to say that he don't have a map yet ? the default map for me should be on the client not the server

lib/devices/RoborockS5.js
```diff
@@ -21,23 +22,22 @@ class RoborockS5 extends Roborock {
...
-            fanSpeeds[MiioVacuum.FAN_SPEEDS.LOW].value = 101;
-            fanSpeeds[MiioVacuum.FAN_SPEEDS.MEDIUM].value = 102;
-            fanSpeeds[MiioVacuum.FAN_SPEEDS.HIGH].value = 103;
-            fanSpeeds[MiioVacuum.FAN_SPEEDS.MAX].value = 104;
-            fanSpeeds[MiioVacuum.FAN_SPEEDS.MOP].value = 105;
+            delete fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.MIN];
+            fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.LOW] = 101;
+            fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.MEDIUM] = 102;
+            fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.HIGH] = 103;
+            fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.MAX] = 104;
```
DeltaEvo, line 40:
```js
const fanSpeeds =
  msg_ver >= 3
    ? {
        [entities.state.attributes.FanSpeedAttribute.VALUE.LOW]: 101,
        [entities.state.attributes.FanSpeedAttribute.VALUE.MEDIUM]: 102,
        [entities.state.attributes.FanSpeedAttribute.VALUE.HIGH]: 103,
        [entities.state.attributes.FanSpeedAttribute.VALUE.MAX]: 104,
      }
    : {
        [entities.state.attributes.FanSpeedAttribute.VALUE.MIN]: 1,
        [entities.state.attributes.FanSpeedAttribute.VALUE.LOW]: 38,
        [entities.state.attributes.FanSpeedAttribute.VALUE.MEDIUM]: 60,
        [entities.state.attributes.FanSpeedAttribute.VALUE.HIGH]: 75,
        [entities.state.attributes.FanSpeedAttribute.VALUE.MAX]: 100,
      };
```

lib/devices/RoborockV1.js
```diff
@@ -1,17 +1,17 @@
...
-        fanSpeeds[MiioVacuum.FAN_SPEEDS.MIN] = {label: "Min", value: 1};
-        fanSpeeds[MiioVacuum.FAN_SPEEDS.LOW] = {label: "Silent", value: 38};
-        fanSpeeds[MiioVacuum.FAN_SPEEDS.MEDIUM] = {label: "Normal", value: 60};
-        fanSpeeds[MiioVacuum.FAN_SPEEDS.HIGH] = {label: "Turbo", value: 75};
-        fanSpeeds[MiioVacuum.FAN_SPEEDS.MAX] = {label: "Max", value: 100};
+        fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.MIN] = 1;
+        fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.LOW] = 38;
+        fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.MEDIUM] = 60;
+        fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.HIGH] = 75;
+        fanSpeeds[entities.state.attributes.FanSpeedAttribute.VALUE.MAX] = 100;
```
DeltaEvo, line 14:
```js
const fanSpeeds = {
  [entities.state.attributes.FanSpeedAttribute.VALUE.MIN]: 1,
  [entities.state.attributes.FanSpeedAttribute.VALUE.LOW]: 38,
  [entities.state.attributes.FanSpeedAttribute.VALUE.MEDIUM]: 60,
  [entities.state.attributes.FanSpeedAttribute.VALUE.HIGH]: 75,
  [entities.state.attributes.FanSpeedAttribute.VALUE.MAX]: 100,
};
```

lib/MqttClient.js
Hypfer: Oh!

I didn't know that this notation exists. Thanks!

Hypfer:
@DeltaEvo 
> for example a Vacuum robot could have multiple batteries

Thats exactly the example I had in mind.

> Why not just insert all the attributes at the construction of the robot class and keep a reference to update them?

Stuff like e.g. the persistent map status isn't available on construction of the robot class which would mean that there would be an attribute with an undefined/null value. 
Another example that comes to my mind would be information related to an extension module (e.g. water tank) which also isn't persistently available.

I do agree though that always creating a new attribute instance feels wasteful. Especially since every update can require up to three array.find calls.

DeltaEvo:
What about something like

```js
constructor() {
    this.persistentMapSettingAttribute = null;
}

parseStatusRes() {
    if (this.persistentMapSetting === null) {
        this.persistentMapSetting = new entities.state.attributes.PersistentMapSettinfAttribute()
     this.robotState.insertAttribute(this.persistentMapSetting)
    }

    this.persistentMapSetting.value = persistentMapSetting
}
```

Or maybe just always create the attribute but with a null/undefined value so in the ui it could be showed but for example disabled (because i see these attributes in the ui like little icons and we  make them gray when no value exist yet, so for example when we have two battery slot but one is removed it's value is null, I think this can apply well to the persistentMapSetting)

(PS: I'm currently starting to write an alternative web ui for Valutedo, I will make a PR soon I think, I don't know yet if this should be added to Valetudo directly [I saw in the 2020+ issue that you wanted it to be separated] but I will make a PR to get feedback and maybe change some backend part :smiley:)

TheLastProject:
I feel LatestCleanupStatisticsAttribute inflates cleanup statistics with a timestamp. What about having the class just be CleanupStatisticsAttribute and containing a timestamp, so that for vacuums which support keeping more statistics than just latest cleanup it would be simpler to implement? Or perhaps even a simple StatisticsAttribute with a timestamp and a type (cleanup, overall, etc.)

Hypfer:
@TheLastProject Personally, I don't like the attribute at all since it isn't state at all.

I just have no idea where to put it instead. It's basically part of the previous cleanup reports but it always references the latest one 🤔

@DeltaEvo The idea was that the UI (or basically everything accessing the state) would check for the presence of an attribute to determine whether or not something is supported. Thats the whole point of that array. Null should be avoided if possible.

lib/devices/MiioVacuum.js
rumpeltux: Then the client would need a way to get the default map (… from the server). So imho there’s little benefit to this.

DeltaEvo: Only if it want to display the default map, i'm developping a alternate client and for example I want to show something else
Without the default map you can know that you don't have a map, with this you don't know because you always have a map

Hypfer:
The ValetudoMap class has been extended to - like the other entities - feature a metaData property which may contain anything.

Known anythings at this moment are 
- `version` - Refering to the Map Format version to handle changes
- `defaultMap` - to indicate that this map is the default map

Furthermore, I think that we'll see something like `mapId` and/or `mapName` to support multiple maps in the future.

The reasoning _why_ these two would be metaData instead of some _real_ attributes of the ValetudoMap class would be that they are in fact metadata and not necessarily required to use the map as a map.

lib/MqttClient.js
```diff
@@ -269,23 +261,49 @@ class MqttClient {
...
+            this.last_ha_state = HA_STATE_MAPPINGS[StatusStateAttribute.value];
+            this.last_state = StatusStateAttribute.value;
+
+            if (StatusStateAttribute.flag !== attributes.StatusStateAttribute.FLAG.NONE) {
+                this.last_state += " " + StatusStateAttribute.flag;
+            }
+        }
+
+        if (areaStatistic) {
+            this.last_cleaned_area = parseFloat((areaStatistic.value / 10000).toFixed(2));
```
alexkn, line 297: We should use the [round](https://github.com/Hypfer/Valetudo/blob/bc29f4eed26516682bb91e3771f3346902680e5b/lib/MqttClient.js#L55) function

lib/MqttClient.js
```diff
@@ -269,23 +261,49 @@ class MqttClient {
...
+        } else {
+            this.last_cleaned_area = 0;
+        }
+
+        if (durationStatistic) {
+            this.last_cleaning_time = parseFloat((durationStatistic.clean_time / 60).toFixed(1));
```
alexkn, line 303: same here

lib/MqttClient.js
```diff
@@ -366,19 +384,55 @@ class MqttClient {
...
+            if (BatteryStateAttribute) {
+                response.battery_level = BatteryStateAttribute.level;
+            }
+
+            if (FanSpeedStateAttribute) {
+                response.fan_speed = FanSpeedStateAttribute.value;
+            }
+
+            if (areaStatistic) {
+                response.cleaned_area = (areaStatistic.value / 10000).toFixed(1);
```
alexkn, line 427: and here

lib/MqttClient.js
```diff
@@ -366,19 +384,55 @@ class MqttClient {
...
+            } else {
+                response.cleaned_area = 0;
+            }
+
+            if (durationStatistic) {
+                response.cleaning_time = (durationStatistic.value / 60).toFixed(1);
```
alexkn, line 433: and here

client/zone/js-modules/map-drawer.js
```diff
@@ -21,9 +21,9 @@ export function MapDrawer() {
 
     /**
      *
-     * @param {Array<Array<number>>} mapData - the data containing the map image (array of pixel offsets and colors)
+     * @param {Array<object>} layers - the data containing the map image (array of pixel offsets)
```
alexkn, line 24: We should use fixed types so that type checking works properly.

lib/entities/SerializableEntity.js
```diff
@@ -0,0 +1,14 @@
+class SerializableEntity {
+    /**
+     *
+     * @param options {object}
+     * @param [options.metaData] {object}
```
alexkn, line 5: I would prefer fixed types instead of that loose metadata objects to support proper type checking.
This also applies to the other classes.

Hypfer: While agree that being statically typed does provide a lot of benefits, I think having static types with dynamic additional properties gives us the best of both worlds.

Nearly everything is still getting type-checked but we also have a lot of flexibility there.

For example this roborock specific quirk:
https://github.com/Hypfer/Valetudo/blob/bc29f4eed26516682bb91e3771f3346902680e5b/lib/devices/Roborock.js#L144-L147
Without dynamic addional properties, this would've required either a new Class which inherits from StatusStateAttribute or some other hack. Neither of those being desirable since it's just for this one vendor-specific quirk.

I also like to think about the metaData property as some kind of staging area which's contents may be refactored into a new sub-class when there is enough demand and/or knowledge about them gathered.

Looking at the PointMapEntity,
https://github.com/Hypfer/Valetudo/blob/bc29f4eed26516682bb91e3771f3346902680e5b/lib/RRMapParser.js#L418-L428

i really don't want to have a `RobotPositionPointMapEntity` which is just the `PointMapEntity` with an `angle` property. However I also don't want to bloat the `PointMapEntity` with an `angle` property since that is specific to this one use-case:
Maybe at some point there are multiple PointMapEntities with an angle. Then, this would get refactored

Interestingly, Shopify does have a similar concept to make it easier for different third-party shopify integrations to interact with each other or add additional information without needing to change a schema:
https://shopify.dev/docs/admin-api/rest/reference/metafield?api[version]=2020-04

lib/devices/MiioVacuum.js
Hypfer: https://github.com/Hypfer/Valetudo/blob/18d435a4b26df301cddedc6234e7c2b2bc669576/lib/entities/map/ValetudoMap.js#L22-L23

This should be sufficient

client/zone/js-modules/map-drawer.js
Hypfer: I agree, however that should happen outside of this PR since it's already insanely huge

client/home.js
```diff
@@ -216,64 +216,58 @@ async function updateHomePage() {
...
 
-        if (res.battery) {
-            batteryStatusText.innerHTML = "Battery: " + res.battery + "%";
-            batteryStatusBar.value = res.battery;
-        }
+        var BatteryAttribute = res.find(e => e.__class === "BatteryAttribute");
+        var StatusAttribute = res.find(e => e.__class === "StatusAttribute");
+        var AreaCleanupStatsAttribute = res.find(e => e.__class === "LatestCleanupStatisticsAttribute" && e.type === "area");
+        var DurationCleanupStatsAttribute = res.find(e => e.__class === "LatestCleanupStatisticsAttribute" && e.type === "duration");
+        var FanSpeedAttribute = res.find(e => e.__class === "FanSpeedAttribute");
```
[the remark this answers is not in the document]
Hypfer: Because the frontend is already broken beyond repair and I'm scared that that might break somewhere

Hypfer:
@rumpeltux Could you share a viomi map file so that I have something to test to port the Viomi MapParser?

PR #549, opened 2020-06-25
add support for client certificate

mqtt-fan:


Hypfer:
https://github.com/mqttjs/MQTT.js/

I can't seem to find that option in the library docs?

mqtt-fan:
@Hypfer you can find it in the [tls client example](https://github.com/mqttjs/MQTT.js/blob/master/examples/tls%20client)/mqttclient.js or in the [pub.js](https://github.com/mqttjs/MQTT.js/blob/master/bin/pub.js) (line 85/89)

Hypfer:
I see.
What about having the certificate in the configuration file? Since it's plain text, we could just use a textArea to paste it.

This would enable the configuration without relying on SSH

mqtt-fan:
I adapted the changes

PR #558, opened 2020-07-14
Smarter room coloring

aa-ko:
This PR attempts to solve the problem that two adjacent rooms can occasionally get assigned the same color in the map, which leads to confusion about the rooms exact boundaries.


client/zone/js-modules/map-color-finder.js
```diff
@@ -0,0 +1,189 @@
+export class MapColorFinder {
```
NKnusperer, line 1: Maybe `FourColorTheoremSolver` is a better class name because it actually describe what it is?
At least this should be noted in the comment that this is a implementation of the theorem.

aa-ko: Agreed. The name is far too generic.

client/zone/js-modules/map-color-finder.js
```diff
@@ -0,0 +1,198 @@
...
+}
+
+class MapAreaGraph {
+    constructor(vertices) {
+        this.vertices = vertices;
+    }
+
+    connectVertices(id1, id2) {
+        if (id1 != undefined && id2 != undefined && id1 != id2) {
+            this.getById(id1)?.appendVertex(id2);
```
Hypfer, line 162: https://caniuse.com/#feat=mdn-javascript_operators_optional_chaining

We probably can't use this syntax yet :(

aa-ko: It's you and the linter against me, so I'll have to comply.

Hypfer:
Man this is completely broken

#560

PR #567, opened 2020-07-25
Fix map coloring

### LINKED ISSUE #566: Web UI map coloring is broken
## The generated map coloring is incorrect if the segment indexes do not start at 0

If the map drawn by the web UI is given segment data in which the segments are not indexed without gaps and starting at 0, the resulting room colors will always be incorrect. Most of the rooms will get assigned the default color, naturally leading to an unreadable map.

## To Reproduce

1. Use any Valetudo version that contains #562 
2. Render any map data with segment indexes not starting at 0 and optionally having gaps in between.

### Screenshots

--

### Valetudo Version

Any version containing #562 

### User-Agent

Mozilla/5.0 (X11; Linux x86_64; rv:78.0) Gecko/20100101 Firefox/78.0

## Expected behavior

The map coloring should be able to transparently map segment indexes in a way that produces a coloring scheme, that is only depending on the actual segment content, ignoring any index or identifier of the segment in the calculation.

aa-ko:
This resolves #566 by changing the segment identification.
Previously, the vertices in the internal graph were indexed by generating a sequence, starting at zero. This PR uses the original segment ID from the attached meta data instead, resulting in correct color resolution when given this ID.
When coloring the graph, the algorithm now sorts the vertices by their number of adjacent vertices in descending order. This is not perfect though and does not *guarantee* that the result will only contain four colors from a mathematical standpoint. However, for all my test cases, this approach did use only four colors.

Additionally, I added a fifth color to the map drawer. Any edge case that is still not resolved by this fix should then fall back to this fifth color.

### DIFF
--- a/client/zone/js-modules/map-color-finder.js
+++ b/client/zone/js-modules/map-color-finder.js
@@ -12,7 +12,7 @@ export class FourColorTheoremSolver {
      */
     constructor(layers, resolution) {
         const prec = Math.floor(resolution);
-        this.stepFunction = function(c) {
+        this.stepFunction = function (c) {
             return c + prec;
         };
         var preparedLayers = this.preprocessLayers(layers);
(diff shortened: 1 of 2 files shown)


aa-ko:
I added a few checks that prevent exceptions from being thrown if the given map data does not contain any segments as advised by @Hypfer .

Issue #551, opened 2020-07-02
Rework REST API


Hypfer:
Proper routes. No routes named after miio commands.
Proper verbs. No PUT request to GET something.

Etc.

Utilizing everything introduced with #544

r1sim:
While documenting the API I noticed some things that could be changed while refactoring the API.

## Combine Endpoints

There are some endpoints that can be unified to one endpoint with multiple possible operations:

| Old Endpoints                                                       | New Endpoint                                                           |
| ------------------------------------------------------------------- | ---------------------------------------------------------------------- |
| `GET /clean_summary` &  `PUT /clean_record`                         | `GET /runs` & `GET /runs/:runId`                                       |
| `DELETE /delete_dnd`, `GET /get_dnd` & `POST /set_dnd`              | `GET /dnd` & `DELETE /dnd` & `PUT /dnd`                                |
| `GET /get_timezone` & `POST /set_timezone`                          | `GET /timezone`& `POST /timezone`                                      |
| `PUT /start_cleaning`,  `PUT /pause_cleaning`, `PUT /stop_cleaning` | `PUT /cleaning_status` (More details in "Bodies that should be added") |
| `PUT /set_sound_volume` & `GET /get_sound_volume`                   | `PUT /sound_volume` & `GET /sound_volume`                              |
| `PUT /test_sound_volume`                                            | `GET /test_sound_volume`                                               |
| `PUT /set_ssh_keys` & `GET /get_ssh_keys`                           | `PUT /ssh_keys` & `GET /ssh_keys`                                      |
| `PUT /find_robot`                                                   | `GET /find_robot`                                                      |
| `PUT /drive_home`                                                   | `GET /drive_home`                                                      |
| `PUT /start_manual_control`                                         | `GET /start_manual_control`                                            |
| `PUT /stop_manual_control`                                          | `GET /stop_manual_control`                                             |
| `GET /get_carpet_mode` & `PUT /set_carpet_mode`                     | `GET /carpet_mode` & `PUT /get_carpet_mode`                            |

## 

## Bodies that should be added



### `PUT /cleaning_status`

This endpoint shall be used to unify all those cleaning action endpoints.

Examples:

**Start Full Cleanup**

```json
{
    "action": "full_cleanup"
}
```

**Pause Cleanup**

```json
{
    "action": "pause_cleanup"
}
```

**Stop Cleanup**

```json
{
    "action": "stop_cleanup"
}
```

**Spot Cleanup**

```json
{
[...]

Issue #379, opened 2019-11-20
[Enhancement] - Set ntp Server via GUI


MrReSc:
## Enhancement
I used the image from dustbuilder. Since you can't specify a local ntp server on the website, I added my local ntp server to the file `/opt/rockrobo/watchdog/ntpserver.conf` and restarted the robot.  This works perfectly. It would be great if you could add an ntp server via GUI.

HHenning0815:
That would be very nice - because my robo has not correct time and I want to Change the NTP Server and have also a button for "sync now" would be fine in this context

pidator:
> That would be very nice - because my robo has not correct time and I want to Change the NTP Server and have also a button for "sync now" would be fine in this context

My robot had a wrong timesetting even with the correct NTP server set. 
You could check the current timezone setting by typing
```
root@rockrobo:~# date
Mon Jan 13 22:28:22 CET 2020
```

If your robot has the wrong timezone, you can set the right one with the command
`root@rockrobo:~# dpkg-reconfigure tzdata`

Hypfer:
Sounds reasonable. I'll await your PR

HHenning0815:
Thanks a lot - that works!

Titanius:
Hello, I have the problem that my time is not right.  I use Valetudo and as a time server as in installation 192.168.178.1. (my Fritzbox) The time server is active in the Fritzbox.  Unfortunately I have the time in the Roborock -1 hour when I query 
$ date
I have also used 
$ dpkg-reconfigure tzdata
to set Europe / Berlin

 Current default time zone: 'Europe / Berlin'
 Local time is now: Wed Mar 4 10:52:20 CET 2020.
 Universal Time is now: Wed Mar 4 09:52:20 UTC 2020. 

I tried this too

echo 'Europe/Berlin' > /mnt/data/rockrobo/timezone
ln -sf /usr/share/zoneinfo/Europe/Berlin /mnt/data/rockrobo/localtime

And in this File

/opt/rockrobo/watchdog/ntpserver.conf

192.168.178.1

why is the time wrong on my Roborock S50?

even a reboot did not help

rumpeltux:
So both UTC and local time are off by an hour each?
If you’re getting a time, but the time is off, chances are that your timeserver is serving a localized time instead. Have you tried using a public timeserver?

Titanius:
So both UTC and local time are off by an hour each? **YES**

[...]

Issue #348, opened 2019-10-20
Gracefully handle connectivity issues in the Web Interface


aimtrainer:
**Describe the bug**

Chrome on android has the option "Add to Home screen". When adding Valetudo to the homescreen and afterwards starting it via the generated shortcut, it is opened as a Web-App, meaning there are no chrome elements visable. When opening Valetudo like this the for the first time it works as expected. When opening it again later, it does not refresh the displayed information. Swiping down does not refresh the page either (as it would normally in chrome) and there's no refresh button either.

**To Reproduce**

Add to Home screen as described above, view info under Home, start cleanup and leave Valetudo. When viewing the Web-App after the Cleanup it does not show the new info about the cleaned area and the cleaning time.

I've also encountered the problem, that I changed the "Persistent Data" setting in the Web-App and after saving it did not show the changed setting in the Web-App (it did when viewing it normally via browser).

**Expected behavior**

When the Web-App is open it should either refresh regularly or at least give the user the opportunity to refresh manually.

**Vacuum Model:** 
RoboRock S50 (Firmware 3.3.9)

**Valetudo Version: ** 
0.4.0

**User-Agent** 
Android Chrome / Web-App

waclaw66:
Seems like websocket issue. Have you tried to kill Chrome before you start shortcut again?

aimtrainer:
Hi Václav,

yea that works, but it's not really a satisfying solution.

Regards Felix

Václav Nováček <notifications@github.com> schrieb am So., 27. Okt. 2019,
11:54:

ottelo9:
I have problems with iOS (iPhone X) and the map not showing if I use the http password login. The map is always grey and a blue loading bard appears on the top of the screen.

Hypfer:
Since pretty much the whole codebase changed recently, please re-check if this is still an issue in [the latest beta](https://github.com/Hypfer/Valetudo/releases/tag/2021.01.0b0) and if yes open a new issue.

Thank you!

Issue #109, opened 2019-02-24
Client-side routing


petabyteboy:
So, there has been some discussion in #90 to move away completely from onsenUI, however I quite like the result it currently gives. What I am really missing is a possibility to have URIs for every page, like /#map, /#zones, ...
Currently, when I reload the page, I always get back to the home page. Having URIs for those pages would solve that problem.

thundergreen:
This is what I also need :)

PR #580, opened 2020-08-05
New UI

DeltaEvo:
This pull request is Draft for a new ui for Valetudo, there is not much yet, I'm opening it to start the discussion.

The current UI is great, but I think we can do better, my goal is to do responsive webapp and not only a mobile webapp, and with onsenUI it's seems harder, and seeing #90 (even if it's closed) you seems to be open for a refactor (?)

And with the new attribute logic I think we can try to do #551 alongside :)

Even if I want to add a desktop ui, I think we need to do it mobile first so I will start with the mobile ui and we will see later

So, first I want to give my personal thought on the current mobile ui, what I like and I want to keep:
- Having the status and the battery on the home page is a good thing
- The Map page is nice, zooming and panning is a must have
- The Dark and White theme is a good feature
- The Settings are well structured and intuitive

What i like less and I want to refactor or discuss about:
- There is a lot of buttons on the home page and some of them seems redundant: play/pause/stop, it look like a remote, and I think we should try to use the space on the home screen to show something else
- The Map page and the Zones pages seems to be redundant for me, this should only be one page
- The Manual Control is not very usable 
- Some of the settings could be presented why some better ui elements, for example a progress bar for the consumables
- The NavBar on the top is hard to access with the hand

This is not a fixed idea, but for now I think I will structure the ui like this:
  - Home Page:
    - The Status of the Robot (State, Battery)
    - A small view of the map to see the progress (+ Area and Time)
    - A Button to Start/Pause cleaning (if Started: Pause, else Start)
    - A Button to Stop/Dock robot (If Started or Paused: Stop, if Stopped: Dock, or maybe just remove stop ?)
    - Small buttons on the home page for navigation (Settings, Map) or a Navbar (I'm not sure yet, will try and see how it goes)
  - Map Page:
    - A Full screen map
    - A Edit button to edit rooms/zones, no-go zones/lines, and all what we need for the Map
 - Remote Control:
    - I don't know yet but I think if I do it It will be a virtual joystick
 - Settings Page:
    - All the current settings

For this new ui I will go with VueJS 2

I want to resolve the following issues: #379 #551 #109 #348 (Making it a PWA ?) and maybe more ?

I don't know yet if this should go in this repo or if @Hypfer you want to separate the UI and Valetudo it can go to another repo

Some resources for inspiration:
- https://dribbble.com/shots/6311150-Robot-Vacuum-Cleaner-App
- https://dribbble.com/shots/7047317-Smart-Home-App-Concept
- https://dribbble.com/shots/3374809-App-for-smart-vacuum-robot
- https://dribbble.com/shots/7729133-Shark-Robot-Map-Controls
- https://dribbble.com/shots/7784185-Shark-Selects
- https://www.youtube.com/watch?v=y1Hrsek0iQM
- https://www.youtube.com/watch?v=vEiUZzoXfPg
- https://www.youtube.com/watch?v=hhQnlEjZ_oA
- https://www.youtube.com/watch?v=bAeXkc8bhqE


Hypfer:
I'll try to work through this line by line and add comments.
If I'm not commenting on something, assume that I agree with you.


> you seems to be open for a refactor (?)

Yes, definitely

>And with the new attribute logic I think we can try to do #551 alongside :)

I'm currently trying out some ideas regarding a capabilities system which would also include #551
There aren't any specifics yet though. That will take a few more days? weeks? Depending on time, mood etc.

The main big difference will be that not all robots provide the same controls. For example an S5 Max might have additional water pump controls which should not be displayed for a regular S5.
There will be standardized capabilities with vendor specific implementations and you should be able to get a list of all supported capabilities for this vacuum

Then you just build UI elements for each of those and display them if they are supported. That's definitely a must for the new UI

>The Dark and White theme is a good feature

Personally, I don't really care about the light theme and would be okay with a default and only dark theme to reduce complexity.

> some of them seems redundant: play/pause/stop

They are semantically different though. A pause is non-recoverable while a stop is. It may be possible to combine play/pause though. Unless I'm missing some edge-case.

> The Manual Control is not very usable

Personally I wouldn't really care if this was removed for the time being. It doesn't add any value imho.

Since the UI should be built around capabilities, the ManualControlCapability (working title) can and should follow at a later point in time.

>if Stopped: Dock, or maybe just remove stop ?

Nah. You might want to completely cancel the cleanup so you can run a new one without sending it back to the dock first.

> separate the UI and Valetudo it can go to another repo

I think that this should be the way to go


> Some resources for inspiration

Great examples! I really like the clean look of those dribbble concepts.
However I think that we should not oversimplify the UI just so that it looks clean.

If there are lots of features there have to be lots of controls. There's no way around that apart from reducing the amount of features. That's not something I'd want to do. Gnome 3 did that :/

DeltaEvo:
> Then you just build UI elements for each of those and display them if they are supported. That's definitely a must for the new UI

Yep that what I was thinking to build the UI with this new capability system

> They are semantically different though. A stop is non-recoverable while a pause is. It may be possible to combine play/pause though. Unless I'm missing some edge-case.

I think also that we can at least combine play/pause, I understand that stop is a different action, but what i was thinking here is to combine some buttons so play with pause, or pause with stop (you pause first and you stop then) or stop with go to dock (your stop first then you go to dock [because when you go to dock this stop the current action] but IDK yet if it's a good idea x) ) to preserve some ui space since we have not a lot on a mobile UI

> I think that this should be the way to go

Ok so I think I will keep the pull request here to keep track of the changes and when it's ready we will see how to handle that (We can still extract all the commits and make another repo with git filter-branch :) )

> Great examples! I really like the clean look of those dribbble concepts.
However I think that we should not oversimplify the UI just so that it looks clean.

> If there are lots of features there have to be lots of controls. There's no way around that apart from reducing the amount of features. That's not something I'd want to do. Gnome 3 did that :/

> Both the first and the last example just waste like 50% of the screen with an image of the vacuum which serves no purpose apart from looking nice.

> So uh. Basically just take Form follows function as the basic design principle for Valetudo and everything should be fine

Yes I agree with that, I don't want to waste 50% of the screen space just for a vacuum cleaner image, I want to add more info on the home page, for me something that is missing is a better view of the progress, when I launch my vacuum robot I keep tracking where it is to try to see if it is stuck or if it moved in an awkward way (for me this means that something is in this path so I need to check) so I want to add a small view of the map and the current state like in the second dribble link

But in every case this pull request is here to discuss about the design so when I will try something I will post a screen or a link to view it here, so we can discuss and iterate on it :)

> So uh. Basically just take `Form follows function` as the basic design principle for Valetudo and everything should be fine
:+1:

Hypfer:
Check out #584 for the updated REST api, capabilities system etc.

tomaszpieczykolan:
Is this PR dead?

DeltaEvo:
I didn't had the time to work on that but it's definitively something that I want to continue :)

Hypfer:
Another thing that might be cool to have in the new UI would be some kind of demo mode where all API calls/responses are mocked.

ccoors:
The capability system provides some abstraction over common functionality, but there are still some minor variations.

> Unfortunately, not all vendors support “stop”. If that’s not the case, stop will perform a pause.
> ~ Valetudo Docs

Maybe the capabilities themselves should "define" the UI. Then every implementation of said capability for every robot could generate the UI for the functions it provides. Maybe a new abstract method `getUIConfig` in the `Capability`-Class that returns a JSON, that the Frontend would parse and build a matching UI? Something like this (just an idea, mixed `BasicControlCapability` and `SpeakerVolumeControlCapability` for demonstration purposes):

```json
{
    "home": {
        "widgets": [
            {
                "type": "button",
                "icon": "play",
                "disabled": "state === cleaning (or maybe some more sophisticated inline scripting code?)",
                "label": "basic_control_clean_start",
                "onClick": [
                    {
                        "type": "put",
                        "payload": {"action": "start"}
                    },
                    {
                        "type": "reload_capability_state"
                    }
                ]
            }, { "type": "button", "...": "..." }
        ]
    },
    "settings": {
        "sound_settings": {
            "title": "sound_settings_title",
            "secondary_title": "sound_settings_title_secondary",
            "icon": "speaker",
            "widgets": [
                {
                    "type": "groupbox",
                    "label": "sound_settings_settings",
                    "widgets": [
                        {
                            "type": "slider",
                            "min": 0,
                            "max": 100,
                            "step": 5,
                            "value": "$this.volume ($this referring to this capability)"
                        },
                        {
                            "type": "button",
                            "label": "sound_settings_save_volume",
                            "onClick": [
                                {
                                    "type": "put",
                                    "payload": {"action": "set_volume", "volume": "$this.volume"}
                                },
                                {
                                    "type": "reload_capability_state"
                                }
                            ]
                        }
                    ]
                }
            ]
        }
    }
}
```

Maybe even go one step further further and invent Valetudo-Capabilities which also generate these UI-Definitions (for stuff like MQTT-Config, that access control config, ...). That would make it much easier to adapt the UI.

I am not sure on how to implement the map stuff though (like the MapSegmentEditCapability). Those might still need to be semi-hardcoded (maybe with custom widgets?) in the Frontend.

csicar:
Some way to help on this PR? What would be needed to complete the PR?

Hypfer:
@csicar take a look at https://github.com/Jomik/Valetudo-UI

csicar:
Is there a plan to merge Valetudi-UI that back into this repo?

Hypfer:
Probably at some point. It's currently in a very early development stage

csicar:
Alright, thanks! I'll contribute to that project then :)

Issue #542, opened 2020-06-20
Rework MQTT Attributes


Hypfer:
The attributes mqtt topic has become crammed with things of very different nature.
It's also not very simple to work with attributes in home assistant in the first place.

Therefore, Valetudo should adopt the way it is done in Tasmota with entities for each attribute grouped by the device specification

Example:
![image](https://user-images.githubusercontent.com/974410/85197829-f1118900-b2e3-11ea-9c4b-71e0b42a50fd.png)

Hypfer:
As mentioned in #568, this should also add the `sw_version` to the device entry

juantxorena:
I'm quite interested in controlling the robot via MQTT. I have no experience with MQTT (other than trying to use this), but I have done some REST APIs. Is there any way I can help with this task, maybe work on some schema for the MQTT API, or something?

duhow:
Adding a list of state attributes based in Roborock S6, added to Home Assistant via Xiaomi Miio:

```yaml
fan_speed_list:
  - Silent
  - Standard
  - Medium
  - Turbo
  - Gentle
battery_level: 100
battery_icon: 'mdi:battery-charging-100'
fan_speed: Turbo
do_not_disturb: 'on'
do_not_disturb_start: '22:00:00'
do_not_disturb_end: '08:00:00'
cleaning_time: 33
cleaned_area: 32
cleaning_count: 463
total_cleaned_area: 10580
total_cleaning_time: 10730
main_brush_left: 120
side_brush_left: 20
filter_left: 82
sensor_dirty_left: 13
status: Charging
clean_start: '2020-10-19T09:44:47'
clean_stop: '2020-10-19T10:18:32'
friendly_name: Roborock
supported_features: 14204
```

PR #584, opened 2020-08-12
Capabilities System

Hypfer:
Another huge infrastructure PR.

## Relevant Changes & Concepts

Here's a quick overview of everything relevant

### Capability Class
https://github.com/Hypfer/Valetudo/blob/c0414363e36f8b3032c9efa66f7806f3f54bb97e/lib/core/capabilities/Capability.js#L3-L13
The base class for everything a robot can do. There's always a generic base class for each feature (e.g. `GoToLocationCapability`) which is extended by multiple vendor-specific implementations (e.g. `RoborockGoToLocationCapability`, `ViomiGoToLocationCapability` etc)

If the base capability class becomes insufficient to implement some new robot, we could do something like `GoToLocationCapabilityV2` to not break backwards compatibility.

See this for an example of a vendor-specific capability implementation
https://github.com/Hypfer/Valetudo/blob/c0414363e36f8b3032c9efa66f7806f3f54bb97e/lib/robots/roborock/capabilities/RoborockGoToLocationCapability.js#L4-L11

### Valetudo Robot Class
https://github.com/Hypfer/Valetudo/blob/c0414363e36f8b3032c9efa66f7806f3f54bb97e/lib/core/ValetudoRobot.js#L5-L13

A thing which wraps its state and capabilities.
Also has its own eventEmitter, so other parts of the application can subscribe to map/status updates on the robot instead of the `Events` class used previously

Capabilities are registered in its constructor like so
https://github.com/Hypfer/Valetudo/blob/c0414363e36f8b3032c9efa66f7806f3f54bb97e/lib/robots/roborock/RoborockV1ValetudoRobot.js#L17-L26

which solves the problem that different robots may support different subsets of all of the vendors possible features which would be hard to implement by simple inheritance.

Furthermore, each robot contains the relevant logic to fetch everything required to communicate with the robot (e.g. path to the token)

### Webserver Changes
The Webserver has been split into different routers. Furthermore, it will only provide routes for capabilities supported by the robot.

For each base capability, there can be a `CapabilityRouter` which provides and encapsulates all routes for the capability in question. The same concept can and should be adopted for the MQTT Interface.

Example:
https://github.com/Hypfer/Valetudo/blob/c0414363e36f8b3032c9efa66f7806f3f54bb97e/lib/webserver/capabilityRouters/ConsumableMonitoringCapabilityRouter.js#L5-L17

The Webserver also provides a meta-endpoint which provides a list of all existing endpoints
https://github.com/Hypfer/Valetudo/blob/c0414363e36f8b3032c9efa66f7806f3f54bb97e/lib/webserver/WebServer.js#L73-L81

In later iterations it would be nice to use something like swagger etc for this.

The generated structure will look like this WIP state:
image

### Configuration Changes

The configuration location has to be provided via an environment variable. This solves the issue that it may be in different locations depending on the platform valetudo is running on. This will require new firmware builds and an updated runscript for all existing robots.

Furthermore, due to the substantial changes to the config format, old configs shall be deleted and everything should be reconfigured by the user. Since `0.6.0` did break all stored zones, there's no point in holding on to that broken config and providing an automatic migration.

### MQTT Changes

The default device name has been changed from `rockrobo` to `robot` to reflect that this isn't just roborock anymore.
This will require documentation updates which is okay, because the whole MQTT structure shall be redone as well as seen in #542 so we can hopefully break everything at once and once only.


alexkn:
I think you should shorten the Capability URLs.
The word "capability" does not have to be included there.

Hypfer:
@alexkn What would be the benefit of shortening those URLs?

While I do agree that they are indeed rather long, I also like the way that they just reuse a property that is already there:
https://github.com/Hypfer/Valetudo/blob/e125dabd27b7392bafd5ccba8dab5adad5a9b72e/lib/webserver/CapabilitiesRouter.js#L31-L35

There could of course be a type to URL mapping but I'd rather not do that just for looks

alexkn:
I think the capabilities system is only internal and shouldn't be reflected in the external api.

Hypfer:
@alexkn

It actually isn't internal since all applications which might interact with the API should build their logic around specific capabilities, so that those can be deprecated/removed/changed at any time.

The idea is that an application fetches all supported capabilities via `/api/v2/robot/capabilities` and then only displays/enables views/functionalities for those specific capabilities

DeltaEvo:
Something that can be good to do with the api update is to replace websocket with [SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) (Server Sent Event).

[SSE](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events) are only one way Server -> Client and simpler than Websockets since it's only http

This seems more adapted for the map update, and you can also use the default http compression 
mechanism on it :)

Also the [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) Api (SSE client on the browser) handle re-connection :)

Hypfer:
This PR is blocking further development of Valetudo. It's been open for so long, I even forgot what is already done and what is still missing :(

Major Todos that have to be done:

[ ] Port Viomi (cc @rumpeltux )
[ ] Modify Client to use new http endpoints

Minor Todos that have to be done:

[ ] Complete MQTT Commands (Zone cleaning, Segment cleaning, Go-To)
[ ] Complete new docs

Major Todos that can be addressed in a different PR:

[ ] Valetudo Timers
[ ] Cleanup history

Major Todos that can be addressed in a different PR:

[ ] Add missing roborock capabilities (e.g. Volume Control)


I might've missed some todos so this comment may be edited

Depau:
I'm filling in this Hoppscotch (FOSS alternative to Postman) requests collection for the v2 API as I'm implementing the capabilities for Viomi, if anyone wants to use it:

- https://gist.github.com/Depau/0a98c369e4bae66ed538948fa0cc5d71
- https://hoppscotch.io

Depau:
I'd like to discuss a few extra capabilities that I'd like to add for the Viomi. I'm asking mainly because I do not know how other vacuums work and I'd like to write code that can be adapted for use with most robots.

First of all, given that
- there's already two "intensity preset" capabilities (fan speed, water usage grade in my viomi PR #623)
- I'd like to add a third one for carpet turbo mode and
- if I understand correctly the new API isn't final yet

I'd like to refactor these 3 capabilities into a single, parametric one that defines a number of "intensities". Each intensity would be selected using an enum, like the different consumables statuses.

With this being said I would like to add the following capabilities:

- Carpet turbo mode
    - as explained earlier, either into this new `IntensityParameterControlCapability` or as a new `CarpetTurboModeControlCapability` which would be identical to fan speed.
    - Viomi supports "off", "medium", "turbo"
- LED control - `LightsControlCapability` or `LEDControlCapability`
    - parametric, with each LED selectable using an enum
    - status + on/off control
- Do not disturb mode  - `DoNotDisturbHoursControlCapability`
    - I do not know how it is implemented in other vacuums, viomi seems to have a "start time" + "end time" (by time I mean clock hour:minute) during which the vacuum will refuse to operate
    - Let me know if roborocks have a similar feature and how it works so that I can make something that can easily map to both makes
- Manual control - `ManualControlCapability`
    - Go left right forward backwards + stop
    - For each movement, optional movement duration if robot supports setting it, viomi does apparently
- Calibration/leveling capability - can't come up with a name
    - viomi allows for sensor calibration, you put it on a flat surface and start the procedure
- Voice control - `VoiceControlCapability`
   - I'm not sure of everything that should be in something like this, but I'd say:
      - Volume control
      - Voice on/off
      - Voice language
      - Custom voice upload
      - Stock voice download from server and installation
   - Another option could be two capabilities, one for mute/unmute + volume control and the other for voice files and language management

I also added a development-only `RawCommandsCapability` in my PR since I use it for testing random payloads at runtime. RFC. 
It is intended to be used only by developers and it must be explicitly enabled in the configuration, under the `debug` object.
I'd say no one should expect it to not have a stable or consistent API. In my viomi implementation it expects a JSON with data to be sent over miIO through sendCommand or sendCloud, and it will send back the response, but I'd say other non-miIO devices can implement it however suits best.

Hypfer:
@Depau 

So the idea of capabilities was, that each of those maps to a specific feature and if it's called BasicControl it will always be basic control.
This way, the WebInterface could build the UI around which capabilities are available and display matching controls etc.
Same for the MQTT Interface of course.
Capabilities should provide a stable interface where you can always be certain, that this model of robot with this capability can do exactly these things. Having to ask the capability what exactly it can or cannot do would go against that.

But of course there can (and probably should) be an abstract `IntensityParameterControlBaseCapability` class which would house all the common logic of these different flavours of IntensityParameters.

Roborock DND is pretty much the same thing as viomi. Start Time, End Time and an `enabled` boolean.

As for manual control, for roborock you do have to enter a "manual control mode" and only then you're able to send commands.
Therefore, it needs to have methods to enter/exit manual control and we most likely need to put the current mode (auto/manual) in the status 🤔


We shouldn't merge too much functionality into the same capability but instead try to split it as much as possible, because it's always possible for a robot to have multiple capabilities but never only half of one.
Therefore, I'd strongly advice against merging all these features into one single `VoiceControlCapability`. Instead, those 5 points would be 5 different capabilities.

The RawCommandsCapability sounds like a great Idea 👍

Depau:
@Hypfer Sounds reasonable, thank you.

So, here's a summary of the capabilities that are already okay to implement (unless you have other comments):

- `CarpetTurboModeControlCapability` similar to fan speed and water grade - I'll try to work out a common abstract base class, I'm going to see what JavaScript's OOM allows for since I'm not very experienced with JS.
- `LEDControlCapability`, method to get a list of LEDs + on/off status, method to turn on/off one LED. LED descriptor will have an interface similar to consumables.
- `DoNotDisturbHoursControlCapability` with start/end times defined as `HH:MM`
- `ManualControlCapability`, forward/backwards/left/right + `enter`/`exit` methods for robots that have a special mode

Here are the ones I changed or new ones for which I'd like a review before working on them:

- **not mentioned earlier** we also need a `SystemTimeControlCapability` to set system date/time and timezone, otherwise DND won't work. I'd actually provide a base class that implements it using generic linux interfaces if `embedded === true`, so that it can still be used if the vacuum software doesn't provide a simple interface.
  - iface: get/set date + time, I'll work out a format but I would say UTC Unix timestamp should get the job done; get/set timezone - I need to figure out what could be a good timezone format since I just checked and viomi doesn't have `/usr/share/zoneinfo` for extracting `Continent/City` timezone data. I will get back on this after investigating.
- `SensorCalibrationCapability`
   - unless all vacuums calibrate the same way, I'd make it provide a list of sensors and a method to calibrate them
- `VolumeControlCapability`, with `getVolume` (%), `setVolume`, `mute`, `unmute`
   - For this capability I would provide an ALSA-based common implementation which robots with no protocol-based interface can use. Are there any vacuums running pulseaudio or other stuff?

Also new/changed capabilities, but with a much lower priority and a lot more prone to be incompatible across vendors - I'd keep these for last:

- `CustomVoicePackManagementCapability`, I'm gonna check the code that implements it for the Roborocks and RE the Viomi equivalent to figure out what could be the best API, voice packs might be highly manufacturer/model dependent
- `VoicePackManagementCapability`, `getAvailableVoices` which returns a JSON containing available packs, language codes and status (enabled, installed, available), `getVoiceDownloadProgress` (which can optionally return indefinite progress if not feasible), `setVoice`

For implementing these caps I'm planning to create a new pull-request targeting `capabilities`. I will also merge this branch into the Viomi PR since I will also implement them for Viomi.

I'm aware that this will make the viomi pull request extremely messy until the other one is merged, but since the Viomi PR will depend on it, it shouldn't be a big issue. I will rebase it and clean it up before it's ready for final review.

Hypfer:
I've greatly reduced the scope of this rewrite further by deciding to comment-out some features that aren't necessary to use Valetudo.
The plan is to have a smaller (pre-)release soon-ish and bring back features on that basis instead of continuing this endlessly-open PR.
Furthermore, the goal of providing a stable API now has been ditched completely, because it's just not feasible.

These things need to be done in order to finally merge this PR:

- Update the UI editor for Zones and Spots so that it doesn't always save an array of everything to work with the new API
- Update the UI editor for SSH Key upload to work with the new API
- Update/Implement/Fix wifi settings
- Update docs (if necessary)
- Provide new deployment instructions and most likely update dustbuilder and other firmware builders (cc @dgiese )
- Testing!

Other than that, I'd say it's pretty much done.

@DeltaEvo I've switched to SSE Map Updates and removed WebSockets entirely. Thanks for the suggestion!

M1cha:
This PR disabled manualcontrol in the UI, why is that the case?

Depau:
> This PR disabled manualcontrol in the UI, why is that the case?

The UI is gone for the time being but you can still use a gamepad ;) https://valetudo.cloud/pages/companion_apps/fun_games.html#remote-control-using-a-gamepad

PR #600, opened 2020-09-18
Various fixes (mostly for Viomi)

rumpeltux:


lib/devices/Viomi.js
```diff
@@ -451,18 +452,15 @@ class Viomi extends MiioVacuum {
             const map = new ViomiMapParser(data).parse();
             const zones = this.configuration.getZones();
             zones.forEach((v, k) => v.user || zones.delete(k));
-            /*
-            Object.entries(map.zones).map(v => {
-                const id = parseInt(v[0]);
-                zones.set(id, {id: id, name: v[1].name, user: false});
-            }); */
+            // TODO: Remove this once Valetudo supports named rooms in the zone clean menu.
```
Hypfer, line 455: That probably won't happen, because Rooms and Zones are two very different things.
Rooms are defined as pixels, while Zones are defined as coordinates.

We should really keep them separated from each other

A separate panel where you can select rooms to clean without the map view would be nice though

rumpeltux: Are they? In viomi maps, rooms are defined by their outline coordinates and a convenience pixel representation is there too. You could also translate zones to pixels.
I think the hard difference is that rooms are mutually exclusive and zones might overlap, but it's not clear to me that we really need the distinction in the "start cleaning something" UI, except that robots probably don’t support mix and matching room & zone cleans (at least viomi doesn’t).

Anyways, thanks for merging for now. I also wouldn’t mind the separate panel.

Issue #591, opened 2020-08-28
Incorrect status and no maps on viomi v7


alceasan:
## Describe the bug
After a clean installation, Valetudo starts and commands work (start, return to home, ...), including MQTT integration, but after starting a cleaning session the status appears always as "docked". The area is being updated while the cleaning is up, and also battery indicator. Also, no map data...
I can start the cleaning from HomeAssistant through MQTT integration, but the problems persists.
## To Reproduce
Install valetudo following the steps on https://valetudo.cloud/pages/installation/viomi.html

### Screenshots

![image](https://user-images.githubusercontent.com/3265883/91560584-5c0dba80-e93a-11ea-9364-8932dbe58534.png)
![image](https://user-images.githubusercontent.com/3265883/91560636-7182e480-e93a-11ea-8f63-88d4cc232420.png)

### Vacuum Model
Viomi v7

### Valetudo Version
0.6.1

### Firmware Version
3.5.3_0047

## Additional context
Not sure if related, but in my vaccum box the model is labeled as STYJ02YM (viomi v8), but after setting up on Mi Home App (on Chinese server, I couldn't do it on Europe servers) and rooting it, the model it apears on /etc/miio/device.conf is viomi.vacuum.v7

Hypfer:
Possibly a duplicate of  #550

alceasan:
Can I provide any relevant logs?

rumpeltux:
1. Make sure /etc/rc.d/S51valetudo has `dest=127.0.0.1`.
2. Same file: the line `iptables  -t nat -A OUTPUT -p tcp --dport 80   -d $host -j DNAT --to-destination $dest:8080`
should change to `$dest:80` (on the robot port is 80 not 8080 as in the development setup).
3. Also check that `/mnt/data/valetudo/config.json` has `map_upload_host` set to `http://127.0.0.1`

Once done, please reboot, start a clean and check again.

alceasan:
I just tried, but it didnn't work. Status continue to show as docked while it's cleaning.

![image](https://user-images.githubusercontent.com/3265883/92247519-9f7ea080-eec7-11ea-91ed-fd4f101eb7cf.png)

alceasan:
Today I changed the deposit from the water tank and mop to the standard waste box and it suddenly started working.

Now the main interface shows correctly the status and map data:
[...]

PR #608, opened 2020-09-30
Fix Viomi status & pause handling that was broken by eb6bece2

rumpeltux:
Addresses broken status handling (#591 & #594)

Don’t merge this as is. We’ll need a solution first how to handle different vendor-specific logic in `buttonStateMap` (see changes in home.js).

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,4 +90,5 @@ This is an outdated list of Valetudo features on Rockrobo vacuums
 * [Valetudo Telegram group](https://t.me/joinchat/AR1z8xOGJQwkApTulyBx1w)
 
 ### Resources
+* [Lovelace Valetudo Map Card](https://github.com/TheLastProject/lovelace-valetudo-map-card) - Map card for Home Assistant
 * [I can't believe it's not Valetudo](https://github.com/Hypfer/ICantBelieveItsNotValetudo) - A companion service for PNG Maps
--- a/client/forbidden-markers-configuration-map.html
+++ b/client/forbidden-markers-configuration-map.html
@@ -16,6 +16,9 @@
         <ons-fab ripple id="forbidden-markers-configuration-add-zone">
             <ons-icon icon="fa-window-close-o"></ons-icon>
         </ons-fab>
+        <ons-fab ripple id="forbidden-markers-configuration-add-mop-zone">
+            <ons-icon icon="fa-stop-circle-o"></ons-icon>
+        </ons-fab>
         <ons-fab ripple id="forbidden-markers-configuration-save">
             <ons-icon icon="fa-save"></ons-icon>
         </ons-fab>
--- a/client/forbidden-markers-configuration-map.js
+++ b/client/forbidden-markers-configuration-map.js
@@ -13,11 +13,17 @@ function markerConfigInit() {
 
 
     const no_go_areas = topPage.data.map.entities.filter(e => e.type === "no_go_area");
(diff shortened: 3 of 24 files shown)


Hypfer:
What about the `resumable` status flag?

https://github.com/Hypfer/Valetudo/blob/ae799a8f4a001120ba2cc34690e1fdef37c07b6d/lib/entities/state/attributes/StatusStateAttribute.js#L48-L55

Can we use that to handle the differences for viomi? It's already used for vendor oddities for roborock as seen here:

https://github.com/Hypfer/Valetudo/blob/ae799a8f4a001120ba2cc34690e1fdef37c07b6d/lib/devices/Roborock.js#L253-L277

rumpeltux:
Sounds good :+1: 

Is there a reason that a state transition from "cleaning" to "start cleaning something else" is disallowed in the UI?
Even if the robot would not directly support this, Valetudo could just issue a "stop" command first before starting the next cleaning, right?

Hypfer:
There's no technical reason for that limitation.

Since that is a "destructive" (the robot forgets its progress) action, I thought that it makes sense to require explicit confirmation to do so

rumpeltux:
To confirm for Roborock:
in state "paused" the stop command should be disabled?
(for viomi it will clear any resumable status and make sure that a subsequent start button press starts a full clean)

Ok I’ll leave the default: stop first then do something else then.

Pausing the "return home" operation doesn’t seem to make too much sense (I also don’t see a resume for that implemented in RR, so I’ll keep the pause button disabled in that case.

rumpeltux:
Also on second thought: I don’t get the `resumable` flag idea, shouldn’t this just be the state `paused`?

In any case I have a hard time mapping the viomi current operation selection into the StatusStateAttribute, because that’s created / updated at other random places. I’ll keep it as is for now.
It may make sense to introduce a separate state attribute that just captures the last started / current operation?

Hypfer:
I just tested it with my S5.
Apparently it is possible to stop from a paused state. This causes a transition into the `idle` state which makes sense.

No idea why I've disabled the stop button for that state.

As for pausing the return home command:
For roborock, it can be resumed by calling the return home command again.
It will say "continue returning to the dock" instead of "going back to the dock" so it does know about that.
However I don't see any difference in behaviour so we might as well just ditch that completely


> Also on second thought: I don’t get the resumable flag idea, shouldn’t this just be the state paused?

While that is true most of the time, for roborock there can also be retuning/resumable and docked/resumable which happens when the robot isn't able to finish the cleanup on one full charge of the battery and then returns to continue later.

Maybe resumable is a bad name for it? 🤔

> It may make sense to introduce a separate state attribute that just captures the last started / current operation?

Sounds reasonable. That would also be a better match for what the resumable flag currently represents.
So we would have the immidiate state of the robot (cleaning, returning, docked, etc) and logic-wise on top of that some kind of task that it is trying to achive?

rumpeltux:
Sorry I don’t have a lot of time to do the refactoring required for capturing the current command into an Attribute.
I’ll keep this on my radar, but want to get the fix in for now to avoid people getting frustrated about broken functionality.
I reinstated the pause functionality for returning-to-dock.

rumpeltux:
I’ll close this for now as this isn’t really mergable to the Capabilities rewrite.

rumpeltux:
Still cc @Depau in case the old reference helps you with your implementation.

PR #609, opened 2020-10-01
Send state to MQTT only if it has been changed

definitio:
Valetudo sends state every 60s even it has same values. It's not necessary because messages sent with retain flag.


Hypfer:
There are recent changes in #584 which would break this and make it a lot harder to implement.

While I do appreciate the effort, I'll have to reject this PR. I also don't think that something like that will come back aftert #584 since those status updates are just so small there's not much to gain there but it introduces new things that may go wrong and/or break something :/

PR #621, opened 2020-11-02
Update supported-dreame-devices.md

i-need-to-tell-you-something:
Title was incorrectly same as supported Roborock devices.


i-need-to-tell-you-something:
Edited this via Github, so I don't know what it did with the last line. Added a newline maybe?

PR #623, opened 2020-11-09
Viomi/capabilities

Depau:
Draft.

The changes as of now provide a stub `MiioValetudoRobot` implementation for Viomi V8 that doesn't crash. Basic status reporting works, capabilities are yet to be reimplemented.

You can consider all changes under `lib/robots/viomi` to be very TODO. Changes in other locations can be reviewed already. I'll add some comments below to clarify why I made them cause it may not be obvious.


lib/miio/RetryWrapper.js
```diff
@@ -93,23 +94,23 @@ class RetryWrapper {
...
-
         const msg = {method: method, params: args};
 
         try {
+            if (!this.miioSocket.stamp.isValid()) {
+                // Trigger retry after performing new handshake
+                // noinspection ExceptionCaughtLocallyJS
+                throw new MiioInvalidStampError();
+            }
+
```
Depau, line 105: Sometimes, if I try to **re**start Valutudo when the robot was already connected, the stamp isn't valid and some code later needs to be reached, but I'm still haven't found it.

This is a workaround, it still prints a bunch of errors later on `newToken.equals()` saying that there's no such method, but eventually it works so I'd keep it as a quick and dirty workaround temporarily until the viomi implementation is in a good shape.

lib/robots/MiioValetudoRobot.js
```diff
@@ -182,31 +182,29 @@ class MiioValetudoRobot extends ValetudoRobot {
...
 
-        } else {
-            return Buffer.from(localSecret, "hex");
+        if (localSecret && localSecret.length >= 16) {
+            return localSecret.slice(0, 16);
         }
+
+        Logger.debug("Invalid localSecret with length " + localSecret.length);
+        return Buffer.alloc(16);
     }
```
Depau, line 209: I refactored this code because the Viomi provides the local secret in `/etc/miio/device.token` and it is not hex-encoded.
This recycles the same code used when run on the vacuum for parsing the value provided in the configuration.

nottherealbatman:
As the Viomi V7 (STYJ02YM) is identical (as far as I know) to Viomi V8 these changes should also work on the V7 then right?

Depau:
> As the Viomi V7 (STYJ02YM) is identical (as far as I know) to Viomi V8 these changes should also work on the V7 then right?

That's possible, you can start testing the REST API and give me some feedback, you can import [this](https://gist.github.com/Depau/0a98c369e4bae66ed538948fa0cc5d71) hoppscotch.io collection and run it locally.

Hypfer:
Why is this closed 🤨

Hypfer:
Anyways. I've just created a patch and applied it to the latest master:

8486f04fdd9176af6e4e7be684bb3eecebaa39e1

Depau:
Alright, great! I'm sorry but I'm currently moving, I'll finish up my work on the Viomi after I'm all settled and I'm caught up with everything, unless someone else steps in before that :)

Issue #611, opened 2020-10-04
Wrong area calculation in Viomi v8


PaskoMoto:
I have realized that the 'Area' calculation in my Viomi V8 is more accurate if AreaCleanupStatsAttribute.value is divided by 1.000 instead of 10.000. i.e:
Changing

`                ("00" + (AreaCleanupStatsAttribute.value / 10000).toFixed(2)).slice(-6) + " m²";`
by

`                ("00" + (AreaCleanupStatsAttribute.value / 1000).toFixed(2)).slice(-6) + " m²";`

in line 305 in the file client/home.js.

I made the change locally and the measurement is accurate to the dimensions of mi apartment, but I am not confident to make a PR since the changes coud probably affect other device models.

Maybe there's another model-specific point in the code where I cuold make the change and test it before issuing a PR?

Best Regards.
PaskoMoto.

schannall:
FWIW: I can confirm; the area seems to be 1/10 of what it should be for me as well.

Hypfer:
While this bug is indeed valid, please don't delete the provided issue template but rather fill out all requested fields.

It makes things much easier.

fhorinek:
The problem is not in client/home.js, but in lib/devices/Viomi.js, during the conversion from m² to cm².
Please look here https://github.com/Hypfer/Valetudo/pull/628#discussion_r528195988

PR #628, opened 2020-11-19
Added function to locate, point clean and zone cleaning

fhorinek:
Hi, I was playing with my new robot viomi.vacuum.v8, and I filled few unimplemented functions:
Robot localisation, point cleaning and zone cleaning.
The functions are tested on my device.

I was using https://github.com/nqkdev/home-assistant-vacuum-styj02ym as reference for commands.

I know that you are currently rewiring capabilities system, but since I have already implemented this, It might be useful as reference for the new internal interface rewrite.

No need to merge.

Thank you for this project!

### DIFF
--- a/lib/robots/viomi/ViomiValetudoRobot.js
+++ b/lib/robots/viomi/ViomiValetudoRobot.js
@@ -74,6 +74,14 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
         this.registerCapability(new capabilities.ViomiCombinedVirtualRestrictionsCapability({
             robot: this
         }));
+        
+        this.registerCapability(new capabilities.ViomiGoToLocationCapability({
+            robot: this
+        }));        
+
+        this.registerCapability(new capabilities.ViomiZoneCleaningCapability({
+            robot: this
+        }));        
     }
 
     setEmbeddedPaths() {
--- a/lib/robots/viomi/capabilities/ViomiGoToLocationCapability.js
+++ b/lib/robots/viomi/capabilities/ViomiGoToLocationCapability.js
@@ -1,0 +1,17 @@
+const GoToLocationCapability = require("../../../core/capabilities/GoToLocationCapability");
+const ViomiMapParser = require("../../../ViomiMapParser");
+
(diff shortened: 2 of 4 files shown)


Depau:
Hi, very useful, thank you! I'll keep this in mind since I'm implementing Viomi support in the capabilities system.

lib/devices/Viomi.js
```diff
@@ -178,7 +178,7 @@ class Viomi extends MiioVacuum {
         if (newState["s_area"]) { //TODO: actually not an attribute. Use Job
             this.robotState.upsertFirstMatchingAttribute(new stateAttrs.LatestCleanupStatisticsAttribute({
                 type: stateAttrs.LatestCleanupStatisticsAttribute.TYPE.AREA,
-                value: newState["s_area"] * 1000 //m² to cm²
+                value: newState["s_area"] * 10000 //m² to cm²
```
Depau, line 181: Isn't this exactly the opposite of what's been reported as a bug here? https://github.com/Hypfer/Valetudo/issues/611#issue-714279647

fhorinek: This is the same issue and same problem.

The vacuum driver in devices/Viomi.js is seting wrong value when converting m² to cm².
PaskoMoto in #611, fix this value during the presentation of this value in client/home.js, but the problem is within the driver.

I am pretty sure that 1m²  = 10000cm²

rumpeltux:
Thank you for the contribution! @fhorinek

I rebased this onto https://github.com/Hypfer/Valetudo/pull/608
fixed the linter warnings and the failing map parser in this new state in https://github.com/rumpeltux/Valetudo/tree/viomi-point-clean

fhorinek:
I think that I broke the references by force-pushing.
And since it is incompatible with current capability system I will leave it here as just reference.
I have more changes here regarding the forbidden areas and virtual walls
https://github.com/rumpeltux/Valetudo/pull/3

Hypfer:
96462ac752b025a3385e31d4ea0cfbea268b207c

^ Virtual Walls & No-Go Areas ported to the capabilities system.

Thx!

fhorinek:
Added capability for zone and goto, they do not work

PR #635, opened 2020-12-27
Add API documentation

r1sim:
Hey,

I was looking at #551 and was trying to figure out the scope of the ticket. Unfortunately there was no API documentation that would help getting an overview of the REST API. 

While I was looking at the API Router, I thought it might be a good Idea to document it in its current state. This might make it easier for people to work with and on the API. I used [Stoplight Studio](https://github.com/stoplightio/studio) wich uses the Open API 3.0 standard. Swagger Editor should also be able to visualize the yaml file.

### DIFF
--- a/lib/webserver/valetudo-api-docs/reference/Valetudo-API.v1.yaml
+++ b/lib/webserver/valetudo-api-docs/reference/Valetudo-API.v1.yaml
@@ -1,0 +1,1984 @@
+openapi: 3.0.0
+info:
+  title: Valetudo API
+  version: '1.0'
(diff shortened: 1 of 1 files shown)


Hypfer:
Thank you for your contribution!

Unfortunately, this API will be removed in the next release so I'm sorry but I'll have to close this one

r1sim:
Oh, damn. I'll check the other branches next time :/

Hypfer:
I've just merged the huge PR which changes the API.

To get an idea on how that looks like now, there's a meta endpoint `/api/v2/` which lists all endpoints

PR #652, opened 2021-01-18
UI and API fix for comsumables

bensweet86:
Capability was in place for consumables, fixes to UI and API to complete the migration to the new codebase.

### DIFF
--- a/client/services/api.service.js
+++ b/client/services/api.service.js
@@ -274,8 +274,8 @@ export class ApiService {
         await this.fetch("PUT", "api/set_lab_status", {lab_status: labStatus});
     }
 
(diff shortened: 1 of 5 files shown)


bensweet86:
Won't build, something else going on with DreameMapParser_spec that is failing but shouldn't be related to these changes

Hypfer:
My bad. I forgot to fix the tests before pushing. You can ignore that

bensweet86:
Ok, I went big to try and fix the problem, will create a new PR shortly to fix this simple one - the rest will come later today when I have a chance. Will also check the docos for the other vacuums to draft up the changes for these as part of the PR although I cannot fully test.

PR #654, opened 2021-01-18
Consumables fix 

Hypfer:
Supersedes #653


Hypfer:
This feature was postponed and I just remembered why.

The return of it in this form works for roborock and roborock only. Every other consumable configuration throws an error due to missing consumables or isn't completely visible in the HTML.
Therefore, the plan was to not touch this and if touched force a rewrite.

Unfortunately, I forgot that which leaves this in a broken state but not broken enough for people to be motivated enough to properly fix it. :/

PR #655, opened 2021-01-19
Roborock/Capability Spot Clean

bensweet86:
Spot cleaning capability for Roborock vacuums added now button on main page functions.  Updated button matrix to disable the stop button when in spot cleaning mode as only the pause function initiates a change in function upon testing with a Gen 1 vacuum.

Function could be replaced at some point with a small fixed zone configuration at a later stage.  Unable to find support for equivalent feature outside of Roborock range.

### DIFF
--- a/client/home.js
+++ b/client/home.js
@@ -296,6 +296,10 @@ async function updateHomePage() {
                     buttonStateMap.spot = false;
                     buttonStateMap.go_to = false;
                     buttonStateMap.zones = false;
+                    switch (StatusStateAttribute.flag) {
+                        case "spot":
+                            buttonStateMap.stop = false;
+                    }
                     break;
                 case "paused":
                     buttonStateMap.stop = false;
--- a/client/services/api.service.js
+++ b/client/services/api.service.js
@@ -55,7 +55,7 @@ export class ApiService {
     }
 
     static async spotClean() {
-        await this.fetch("PUT", "api/spot_clean");
+        await this.fetch("PUT", "api/v2/robot/capabilities/SpotCleaningCapability");
     }
 
     /**
--- a/lib/core/capabilities/SpotCleaningCapability.js
+++ b/lib/core/capabilities/SpotCleaningCapability.js
@@ -1,0 +1,20 @@
+const Capability = require("./Capability");
+const NotImplementedError = require("../NotImplementedError");
+
+class SpotCleaningCapability extends Capability {
+    /**
+     * @abstract
+     * @returns {Promise<void>}
+     */
+    async spotClean() {
+        throw new NotImplementedError();
+    }
+
+    getType() {
+        return SpotCleaningCapability.TYPE;
+    }
+}
+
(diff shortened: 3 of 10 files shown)


Hypfer:
There seem to be quite some differences regarding Spot Cleaning across the different vendors which is (I assume) caused by the fact that only roborock seems to have a GoTo feature.

Apart from Roborock, both viomi as well as dreame require x/y coordinates of the spot they should clean.
According to some review site, the 360 S6 even supports multiple iterations for spot cleaning


<br/>
Unless I'm missing something, the whole feature feels like a worse version of zone cleaning which raises the question if it should be implemented at all, since I can't imagine a situation where someone would use Valetudo to rather do a spot cleanup instead of a zoned cleanup.<br/>
The main use-case of it I'm seeing would be people pressing the physical spot cleanup button on the robot instead of opening up the App but that works without Valetudo support<br/><br/>


I'm sorry that this thought came _after_ you've already finished implementing the PR :(

bensweet86:
Depends how you look at it.  Zone cleaning works well for routine cleaning or where you want to define rooms, etc.  If you use spot cleaning for those scenarios then it absolutely is a worse use case.  From a practicality point of view to pick an area on the map that you know needs a routine clean should in all reality use a zone.  Spot cleaning has its uses to go to a point and spot clean (2x instructions for the Roborock, 1x instruction for others) as if you use Valetudo the vacuum knows where the dock is and how to get back.  With the Gen1's at least there is no guarantee if you pickup the vacuum and pit it somewhere that it will figure it way back to the dock reliably from what I've found.  In this situation one could use a zone or spot to similar effect with spot having the edge if you need the simpler workflow to go and have another parse at an area instead of drawing a region, etc.

For me there is a use case, albeit limited - in any case the UI, etc needs to be capable of stopping the vacuum, or sending the vacuum to a point (both with individual use cases separate to zones alone) so to throw in the spot cleaning capability while I was at it was logical.

For the other vacuums that require an x,y it would be simple enough to do one of the following:

- In Valetudo the workflow for Roborock is select a point on the map, hit spot clean or simply hit spot clean and it will by default do the current zone (provided the vacuum isn't docked)
- For non roborock vacuums the workflow isn't different, except the internal x/y from Valetudo would be the missing context and is readily available.

API endpoint would be no different, only change would be in the implementation of the capability to use those internal X/Y's.

Hypfer:
But how would it be better to use the map to drive the robot to a marker and trigger a spot cleanup, when one could also draw a temporary zone on the map and start a cleanup of that?

There's of course manually carrying the robot somewhere and triggering a spot cleaning there, but for that using the spot clean hardware buttons on the robot itself would make much more sense than opening up Valetudo

bensweet86:
Both ways would achieve a similar / same outcome at the end of the day.  Practically with the Gen1 and potentially because it didn't have zones originally we found goto / spot a good option to have another go at an areas the kids had made a real mess of.

PR #656, opened 2021-01-19
Roborock/Capability Do Not Disturb

bensweet86:
Do not disturb capability for Roborock vacuums, brought back on the relevant WebUI for now given that the rest of the timers page should be replaced with the internal job logic.  Do not disturb could be added to the job functionality too, main reason to bring back separately is the need to set DND on board the vacuum so it doesn't resume a clean or perform its own functions within the DND timeframe outside of the control of Valetudo.  Includes

- Capability router with get / post (multiple dnd support later on?) and delete routes
- Delete API requires delete in body to confirm operation
- Set DND API takes new structured json object with start / end times each with hour / minute parameters
- State attribute for use later on with Jobs holding the status including enabled / disabled and time in similarly structured json object to what the set DND requires


Hypfer:
I've done some research regarding the situation of other brands (namely Viomi, Dreame, 360, Ecovacs) and the DND feature consists of exactly these parameters for all of them:

- State (Enabled/Disabled)
- StartTime
  - Hour
  - Minute
- EndTime
  - Hour
  - Minute

Therefore, multiple DNDs shouldn't be something supported by this capability.
If there's ever something like that, I'd say that we'd rather add a new capability for it.

lib/robots/roborock/capabilities/RoborockDoNotDisturbCapability.js
```diff
@@ -0,0 +1,58 @@
...
+                    hour: data[0].start_hour,
+                    minute: data[0].start_minute
+                },
+                end: {
+                    hour: data[0].end_hour,
+                    minute: data[0].end_minute
+                }
+            })];
+
+        dnd.forEach(d => this.robot.state.upsertFirstMatchingAttribute(d));
```
Hypfer, line 28: DnD shouldn't be part of the RobotState but instead queried manually when it is required.

The RobotState should exclusively contain state information and no configurations

lib/entities/state/attributes/DoNotDisturbAttribute.js
```diff
@@ -0,0 +1,23 @@
+const StateAttribute = require("./StateAttribute");
+
+class DoNotDisturbAttribute extends StateAttribute {
```
Hypfer, line 3: This shouldn't be a State Attribute but instead called `ValetudoDNDConfiguration` which extends `SerializableEntity` and would be found in `lib/entities/core`.

There you can also find the `ValetudoWifiConfiguration` class, which does the same thing but for wifi

lib/core/capabilities/DoNotDisturbCapability.js
```diff
@@ -0,0 +1,39 @@
...
+const NotImplementedError = require("../NotImplementedError");
+
+class DoNotDisturbCapability extends Capability {
+    /**
+     * This function polls the current consumables state and stores the attributes in our robotState
+     *
+     * @abstract
+     * @returns {Promise<Array<import("../../entities/state/attributes/DoNotDisturbAttribute")>>}
+     */
+    async getDnd() {
```
Hypfer, line 11: getConfiguration which returns a ValetudoDNDConfiguration

lib/core/capabilities/DoNotDisturbCapability.js
```diff
@@ -0,0 +1,39 @@
...
+        throw new NotImplementedError();
+    }
+
+    /**
+     * @abstract
+     * @returns {Promise<void>}
+     */
+    async deleteDnd() {
```
Hypfer, line 19: Remove this

lib/core/capabilities/DoNotDisturbCapability.js
```diff
@@ -0,0 +1,39 @@
...
+        throw new NotImplementedError();
+    }
+
+    /**
+     * @abstract
+     * @param {object} preset
+     * @returns {Promise<void>}
+     */
+    async setDnd(preset) {
```
Hypfer, line 28: setConfiguration which takes a ValetudoDNDConfiguration

Hypfer (changes requested):

Hypfer:
Looks good so far. 👍
I've added a few remarks where appropriate

bensweet86:
Had a similar thought process myself where to put this. The DND enabled is more a state attribute to me and parameters as you say should go elsewhere.

The state of DND is important if the Jobs functionality needs to understand whether the schedule should be overridden or not.

I was thinking the next iteration would see the DND not only relate to core OEM functionality but also extend to a Valetudo state where it understands for a schedule whether to even try and initiate and activity or not as there could be additional use cases outside of the original vacuum functionality such as:
- Clean at night with no voice / volume at 0
- Option to resume a clean even if this is within a DND period

Thoughts?

Hypfer:
At least for roborock it's (afaik) impossible to override the DND settings but even if it was possible that's far beyond the scope the scheduling in Valetudo should have.

In a perfect world, there would be no timers inside valetudo whatsoever since it would just provide the generic interface for more sophisticated home automation software to do the scheduling.
There are a few valid use-cases where having to install Home Assistant on a different host just to get a very basic schedule running would be overkill though. (e.g. the vacuum robot at your (grand-)parents house etc)
Therefore there doesn't seem to be a way around having a basic timer functionality to cater to that needs.
However everything beyond that very basic scope should be implemented elsewhere.

A while ago I've suggested that people could add their sophisticated automation configurations to the docs so that other users may simply copy-paste those into their own home automation setups.
That never happend though :(


It would be possible to either poll the DnD config via HTTP or publish it to another mqtt topic without having it as a state attribute, however I don't really see the point since there's nothing that could be done with that information due to the firmware handling it internally. You can tell it to continue but it will iirc do nothing or start a new cleanup which both isn't the intended action

bensweet86:
Don't disagree with those comments, from reading the Telegram thread I thought the intention was to add cron timer functionality within Valetudo to initiate schedules instead of relying on the vacuum from starting a schedule itself.  Example would be in the Gen1's it can start a cleaning job on a schedule but not a zoned / goto / spot clean.  In this instance there would be an opportunity to not only build on the scheduling logic but include DND, etc as enhancements too.

In any case will push though an updated PR to move DND away from the state attributes when I get a change later tonight.

lib/core/capabilities/DoNotDisturbCapability.js
bensweet86: deleteDND retained but renamed for consistency to deleteDNDConfiguration.  The only way to disable DND on Roborocks is to send close_dnd_timer as set_dnd_timer only accepts start hour / minute and end hour / minute.

Hypfer: Hm but what about looking at the `enabled` property of the `ValetudoDNDConfiguration` and selecting the correct miio command accordingly?

Hypfer:
Unfortunately, I can't push my proposed changes to your branch. Therefore I've opened #659

PR #661, opened 2021-01-22
Roborock/Capability Carpet Mode

bensweet86:
Carpet mode capability for Roborock vacuums. Includes

- Capability router with get / put for managing settings & enabled status
- ValetudoCarpetModeConfiguration entity to store configuration through serializable entity
- Capability across roborock vacuums, tested with Gen1


Hypfer:
As mentioned in the Telegram group, I'd like to omit the additional parameters and instead just have a simple boolean toggle like the `PersistentMapControlCapability`, since those parameters aren't documented and most vendors don't even allow to edit them at all

bensweet86:
I tested the capability of Roborock to handle only the enabled status, issue is that set_carpet_mode requires the other parameters as if enabled is the only parameter toggled, the other parameters such as stall time and current low / high / integral are set to 0 when not supplied.
This implements the complete functionality, otherwise would require the workaround as part of the config application to the vacuum:

- get_carpet_mode to get the list of these parameters
- set_carpet_mode would initially need to get_carpet_mode to get the current values, overwriting the enabled status with the UI value

If we still need to include get as part of the set process why not have the capability to see and potentially edit these parameters?
It wouldn't take much to allow a vacuum with the parameters to toggle vs another one with just an enabled flag to be supported with parts of the API endpoints optional.

Hypfer:
> set_carpet_mode requires the other parameters

I'd simply hard-code the default values for set_carpet_mode since they're default for a reason

> why not have the capability to see and potentially edit these parameters?

Because it's a generic capability + capabilityRouter and it makes no sense for it to take roborock-specific arguments.

The overall goal of the rewrite was to be less vendor-specific which also means that not everything that would in theory be possible will also be implemented in Valetudo.<br/>
The whole idea being that you'll get mostly the same experience, no matter what robot Valetudo is running on.
(Minus polyfilling of course since that can become very complicated very fast)

If there are vendor-specific configuration options, reasonable defaults which fit 80% of the users should be used.


<br/><br/>
I appreciate your contributions, however maybe you'd be better off using Valetudo RE since that project is focussed more on supporting more vendor-specific details.

It also already supports map backup&restore which is very useful for the roborock v1 but doesn't really make sense to implement in Valetudo due to the market having moved forward from that and it being highly unlikely that we'll see non-persistent-map vacuum robots being released in the future

bensweet86:
Don't get me wrong, I have never had a need to customise these parameters so driven to include.  If I take the parameters hardcoded in the UI as the defaults, these were different to the parameters that were in my Gen1 out of the box.  Perhaps then the best option is to - as part of the Roborock capability - would be to stick with the get / set instead of hardcoding to ensure we aren't changing a value unintentionally.

Hypfer:
Does the firmware reject the command if it is sent without parameters? 🤔

Another approach could be to extend the Capability's contructor to accept the default values for this model similar to this:

https://github.com/Hypfer/Valetudo/blob/4b3ed974d064b788f470c9439a719a3f3ed95cb9/lib/robots/dreame/capabilities/DreameBasicControlCapability.js#L3-L41

and then providing those inside the Vacuum implementation like that

https://github.com/Hypfer/Valetudo/blob/4b3ed974d064b788f470c9439a719a3f3ed95cb9/lib/robots/dreame/DreameD9ValetudoRobot.js#L114-L134

bensweet86:
Yes, I'm going down a path similar to what you noted above.  Thanks for the pointer.

The Roborock API in this instance will actually accept the request if those additional parameters haven't been supplied, which is are then defaulted to 0.  Only having a Gen1 and knowing this actually isn't supported I cannot say if this is normal or not, presume it is so we will have to handle it accordingly by requesting the vacuum state, toggling enabled then sending back the modified configuration in full.

bensweet86:
@Hypfer, changes updated.  The status of carpet mode isn't emitted in the base Roborock state attributes so functionality reverted back to basis on / off toggle with no entity.  What is the position for these state attributes such as this (not yet added) and consumable state attribute (exists) where the state is only updated currently when the settings section loads the page? i.e. the state isn't periodically updated through the ValetudoRobot.pollState() function.

bensweet86:
https://github.com/Hypfer/Valetudo/blob/4b3ed974d064b788f470c9439a719a3f3ed95cb9/lib/robots/roborock/RoborockValetudoRobot.js#L221-L227

bensweet86:
https://github.com/Hypfer/Valetudo/blob/4b3ed974d064b788f470c9439a719a3f3ed95cb9/lib/robots/roborock/capabilities/RoborockConsumableMonitoringCapability.js#L61-L66

PR #663, opened 2021-01-24
Roborock/Capability Sound Control

bensweet86:
Sound capability for Roborock's, implements additional "test_speaker" in SpeakerVolumeControl capability.

Disabled voice upload functionality in UI, additional work to validate how this works across other robots.  Looks like implementation across Viomi's is substantially different.


Hypfer:
While looking at this I've noticed that the existing SpeakerVolumeControlCapability interface didn't really fit what's in the market, since there's no robot with mute support and nor does it make much sense, since it's the equivalent of setting the volume to 0.

It gets worse when unmuting comes into play, because for that one would need to know the previous volume setting which we don't have.

It also seems like there are some robots which don't support testing the volume which is why I moved that to a different capability.

I've adapted your changes to the changes of the master branch here 4a47939ebea220e19ed9da3b4427c53da429e700

Thanks!

bensweet86:
Looks good, abbreviated mute functionality was there in the interim to avoid the capability not implemented error if mute was implemented on others.  At least now its aligned to not have the mute function that was in the original capabilities branch.

PR #667, opened 2021-01-29
A bunch of fixes from trying to run viomi-v7 in local development

rumpeltux:


lib/webserver/WebServer.js
```diff
@@ -97,15 +95,16 @@ class WebServer {
 
              This is a hack for miio vacuums with a recent miio_client
 
+             Dreame doesn’t support iptable port redirects.
              To properly spoof the http_dns request, we need to have this route on port 80 instead of the
              miio-implementation specific second webserver on 8087 :/
          */
-        if (this.robot instanceof MiioValetudoRobot) {
+        if (this.robot instanceof DreameValetudoRobot) {
```
Hypfer, line 102: We also neeed this for newer roborock vacuums and basically everything that uses the miio_client version 3.5.8 and up

rumpeltux: But Roborock is able to use the iptables workaround, right? Today I think that’s still better than patching the binary because it allows us to do local development.

Probably we just need to edit the `rc.local` file to change the redirect for port 80 to 8079.
And maybe fix `embeddedDummycloudIp = "127.0.0.1"` in `RoborockValetudoRobot.js`

Hypfer: Not all/not effortlessly. It has been changed to the loopback approach in the latest dustbuilder release since it's a lot cleaner for regular deployments imo

rumpeltux: I see. So maybe we could keep the GSLB part on the device and then have the on-device valetudo server send the development host IP in the response. We might then get away with not requiring iptables even for local development anymore.
And if we’re already patching miio_client we can also patch the /gslb destination port I guess.

Hypfer: oooh thats a great idea!

Hypfer: Patching the port might be much harder though. For the IP's it's just

```
sed -i -E 's/110.43.0.83/203.0.113.1/g' miio_client
sed -i -E 's/110.43.0.85/203.0.113.1/g' miio_client
```
and even that wouldn't necessarily be required since we could also add two more loopback aliases with those two IPs.
Patching it does prevent accidential cloud connections on misconfigured networks though

rumpeltux: I took a brief look: the port is hardcoded in the binary. So we would a binary patch and the patch would be slightly different for each miio_client.
E.g. the viomi one manually calls htons() on the port and roborock has this inlined. With different versions register allocations might change, so we’d need to update the patch for new versions too… So probably not worth the effort.

rumpeltux: We probably also can do:

    sed -i 's/110.43.0.8./127.00.00.1/g' miio_client

and avoid adding the additional local interface alltogether?

dgiese: yeah, that might actually work

PR #669, opened 2021-01-30
Roborock/Clean history capability

bensweet86:
Implementation of the cleaning summary and cleaning history using the Jobs structures.

- No longer tied to the Roborock timezone, based on UTC and UI converts to locale
- New JobHistoryCapability that can provide a summary of many jobs or individual records
- Minor change to Job definition to include Job count and declare variables
- Additional job type to handle summary information as distinct from detailed records

Also includes brining back the UI for cleaning history and bringing back lifetime summary statistics to the consumables page.


bensweet86:
@Hypfer, I have another addition to my fork where the Job / Clean History has the ability to request and display the corresponding map.  Do you have any intention of merging this PR as the map addition extends this PR? [https://github.com/bensweet86/Valetudo/commit/e2a5ed3635415cf4e6c7d77146ade9634508bd4c](url) is the main commit with the bulk of these changes?  I can either extend this PR to include the map changes or create another PR once this one is merged..

Hypfer:
I've spent a lot of time changing the codebase so that it isn't absolutely painful to work with.
Now, I see this:
https://github.com/bensweet86/Valetudo/commit/54a786e4b38fc7eef2578c4c835023f7e766c594#diff-3e0f58a142ee33bbbb9476f3bd4d778e402577696c2a916f36a7c746893558d0R311-R347

A magic flag set somewhere that influences the flow elsewhere.
What a nightmare to debug if something goes wrong.

I get that it's frustrating to not get an answer for a whole month but as I said, I currently don't have the time to properly think this topic through. I'm sure that your solution works, but that's just not the goal for this project.

Again:
Just use RE.
It has all the features that you want.

There is no point in hacking those into the new and improved codebase.
_If_ they ever appear, it will at least take 5x the time to implement them, because it should be done right and not introduce another boatload of complexity.

That's also the point of the capabilities system. It should be very easy to implement an existing capability for a new robot. It should be incredibly hard to think of a new core capability.
If it's not, there are definitely corners that are being cut and that just doesn't make sense for something that defines the infrastructure.

bensweet86:
My comment was more related to whether there is value in this PR, if not I'll close it.  Likewise if showing the map isn't core there is no point raising a PR.  The code you pointed out wasn't intended for a PR merge - it was a hack to make a vacuum running RE (we have 2, specific use case for RE, otherwise they will soon both be hooked into HA) which is why it made it into my forks main branch.  Will start a discussion thread for the JobHistory capability to get it out of this PR.

PR #682, opened 2021-02-07
Feature: pre-commit

ccoors:
[pre-commit](https://pre-commit.com) is used by many projects to make it easy to enforce code style and avoid other common problems.

The suggested configuration runs [commitlint](https://commitlint.js.org/) to enforce conventional commits, checks json and yaml files for syntax errors, deletes trailing whitespaces, adds a line break to the end of every text file, removes Unicode-BOMs, runs eslint, ... (see `.pre-commit-config.yaml`)

The integration into GitHub Action is based on https://github.com/marketplace/actions/pre-commit .


Hypfer:
So this is run before things are pushed to github, right?

Does it work on windows as well?

ccoors:
That's the plan :) Just install it with `pre-commit install`.

Yeah, it works on windows, though installation is a bit more complicated (as usual ;)). But it's completely optional, as the checks will also run in the PR.

Hypfer:
While I really appreciate the Idea, I'm a bit sceptical regarding pre-commit since it introduces python as a completely new (albeit optional) dependency to the project.

Furthermore, I'm not a fan of their philosophy

> We believe that you should always use the best industry standard linters. Some of the best linters are written in languages that you do not use in your project or have installed on your machine. For example scss-lint is a linter for SCSS written in Ruby. If you’re writing a project in node you should be able to use scss-lint as a pre-commit hook without adding a Gemfile to your project or understanding how to get scss-lint installed.

Understanding what I'm doing beats using "industry standard" tools every time imo.

>  If one of your developers doesn’t have node installed but modifies a JavaScript file, pre-commit automatically handles downloading and building node to run eslint without root.

This goes far beyond what I'd expect my tools to do.


As I said I really appreciate the input but this tool unfortunately doesn't fit this project

ccoors:
That's fine, I understand your concerns and they are definitely valid. Thanks for taking your time and looking into it!

PR #690, opened 2021-02-15
Add Valetudo log to UI

ccoors:
We talked about this :)

This adds support for getting the Log in the UI and setting the log level:

image

I am not sure about the default location of the logfile: `path.join(os.tmpdir(), "valetudo.log");` - A better location might be in `/var/log/...`? Maybe depends on the robot. A default location of `/dev/null` would also work, though then the UI Log would not work ("Empty logfile" message).

Logging is directed to both that file and `console.log`. Not sure about removing `console.log`.

### DIFF
--- a/lib/Logger.js
+++ b/lib/Logger.js
@@ -1,3 +1,4 @@
+const {EventEmitter} = require("events");
 const fs = require("fs");
 const util = require("util");
 
@@ -8,13 +9,19 @@ const LogFileOptions = Object.freeze({
 let LogLevel = 0;
 let LogFilename = "/dev/null";
 let LogFile = fs.createWriteStream(LogFilename, LogFileOptions);
+let LogEventEmitter = new EventEmitter();
 
 const LogLevels = Object.freeze({
-    "trace": -2,
-    "debug": -1,
-    "info": 0,
-    "warn": 1,
-    "error": 2
+    // eslint-disable-next-line no-console
+    "trace": {"level": -2, "callback": console.debug},
+    // eslint-disable-next-line no-console
+    "debug": {"level": -1, "callback": console.debug},
+    // eslint-disable-next-line no-console
+    "info": {"level": 0, "callback": console.info},
+    // eslint-disable-next-line no-console
+    "warn": {"level": 1, "callback": console.warn},
+    // eslint-disable-next-line no-console
+    "error": {"level": 2, "callback": console.error},
 });
 
 const BuildPrefix = (LogLevel) => {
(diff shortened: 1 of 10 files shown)


Hypfer:
Would it be possible to be able to use stdout?

For development on my local machine, I really don't want to have to look into another logfile. Running Valetudo interactively is much easier in that scenario

lib/Logger.js
```diff
@@ -20,72 +29,83 @@ class Logger {
...
+    }
+    static get LogFile() {
+        return LogFilename;
+    }
+
+    static log(level, ...args) {
+        if (LogLevel <= LogLevels[level]) {
+            const logPrefix = BuildPrefix(level.toUpperCase());
+            // eslint-disable-next-line no-console
+            console.log(logPrefix, ...args);
```
Hypfer, line 59: Before, the code was using `console.log`, `console.warn`, `console.error` etc which caused the log to be color-coded on supported setups without any additional logic as well as some stuff getting logged to stderr instead of stdout.

The latter wasn't utilized though.

The downside of using those methods was that the log was sometimes not chronologically sorted for some reason.

lib/Logger.js
```diff
@@ -20,72 +29,83 @@ class Logger {
...
+            LogFile.write([logPrefix, ...args].map(arg => {
+                if (typeof arg === "string") {
+                    return arg;
+                }
+                return util.inspect(arg);
+            }).join(" "));
+            LogFile.write("\n");
+        }
```
Hypfer, line 67: While we're at it, it might also make sense to think about logging to mqtt maybe?

lib/webserver/ValetudoRouter.js
```diff
@@ -29,6 +29,29 @@ class ValetudoRouter {
...
+                current: Logger.LogLevel,
+                presets: Object.keys(Logger.LogLevels)
+            });
+        });
+
+        this.router.put("/log/level", (req, res) => {
+            if (req.body && req.body.level && typeof req.body.level === "string") {
+                Logger.LogLevel = req.body.level;
+                const loggingConfig = this.config.get("logging");
+                this.config.set("logging", {
```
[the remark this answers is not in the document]
ccoors: Documented and done

lib/Logger.js
ccoors: Sure. Added that, not an MQTT-Expert though, so please have a close look.

lib/mqtt/MqttClient.js
```diff
@@ -52,6 +52,9 @@ class MqttClient {
             }
         });
 
+        Logger.onLogMessage((line) => {
```
Hypfer, line 55: I really like the .onLogMessage, however I wasn't really expecting that logging to mqtt would be implemented straight away. Sorry for being unclear there

Due to issues like secrets leaking somewhere, I'd propose commenting out the mqtt part for now

PR #691, opened 2021-02-16
Do not log twice if logfile is the same as stdout redirection

ccoors:
This prevents duplicate log entries if stdlog is redirected to the same file as the actual logfile.

Implementation of `Tools.ARE_SAME_FILES` is similar to https://github.com/python/cpython/blob/master/Lib/genericpath.py#L85


Hypfer:
To my surprise, this doesn't break development on windows since the try/catch also catches `/proc` not existing at all.

Neat

PR #698, opened 2021-02-17
Do not cache API responses

ccoors:
Sometimes the browser cache seems to not return even the cached responses. As the payload of API calls is fairly small, turning off caching should not be a performance issue.

### DIFF
--- a/client/services/api.service.js
+++ b/client/services/api.service.js
@@ -12,7 +12,8 @@ export class ApiService {
             headers: {
                 "Content-Type": "application/json"
             },
-            body: JSON.stringify(body)
+            body: JSON.stringify(body),
+            cache: "no-store"
         });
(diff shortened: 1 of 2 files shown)


Hypfer:
> Sometimes the browser cache seems to not return even the cached responses. 

Are you sure? That might be the rate-limit introduced in 8fd53a5237bc847708b7ee631da4afbd8fe7b950

ccoors:
image

Yeah. And I would even increase the rate limiting, 5/min seems a bit low. It really isn't that much load to send.. a file. Basically what a webserver does best ;)

ccoors:
To reproduce: Visit the info page, the log loads. Leave the page and open it again. The log is gone and can not even be retrieved with the button.

PR #699, opened 2021-02-18
Add functionality to run S4 Max

flacjacket:
Create the configuration for the S4 Max.  This robot adds the multimap persistence functionality to the S4 (basically the same as the S5 Max without the mop).  This has been tested on my current S4 Max and works correctly.

### DIFF
--- a/lib/core/ValetudoRobotFactory.js
+++ b/lib/core/ValetudoRobotFactory.js
@@ -60,6 +60,7 @@ const Robots = {
     "RoborockS6ValetudoRobot": require("../robots/roborock/RoborockS6ValetudoRobot"),
     "RoborockS4ValetudoRobot": require("../robots/roborock/RoborockS4ValetudoRobot"),
     "RoborockS6PureValetudoRobot": require("../robots/roborock/RoborockS6PureValetudoRobot"),
+    "RoborockS4MaxValetudoRobot": require("../robots/roborock/RoborockS4MaxValetudoRobot"),
     "RoborockS5MaxValetudoRobot": require("../robots/roborock/RoborockS5MaxValetudoRobot"),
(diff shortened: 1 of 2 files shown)


Hypfer:
What about the flow of unhandled properties_changed messages you've reported?

flacjacket:
Right, I wasn't sure what those messages meant or what needed to be done about them. After putting this up, I saw there was some logic that looked similar for the Dreame D9, and I was able to find the miot spec page for the S4 Max. Doing some more testing after I got it into home assistant, it looks like I need to handle those messages to do things like update state. The state in HA wouldn't update on its own, but would update if I opened the Valetudo web UI, which I'm guessing forces an update.

If you have any other thoughts or pointers to resources that would help, that would be appreciated. I'll report back when I am at least able to run a few different cleaning tasks and keep the clean log of those messages.

Hypfer:
There are tiny bits of information about the miot protocol spread all over the internet. Unfortunately, it's all chinese

For example, this table contains a definition of services, properties, actions and events which map to `siid`, `piid`, `aiid` and `eiid`

https://github.com/MiEcosystem/miio_open/blob/master/arduino/Arduino_Firmware.md#5-miot-spec%E8%AE%BE%E5%A4%87%E5%8A%9F%E8%83%BD%E5%AE%9A%E4%B9%89

There are also the mi IoT docs here: https://iot.mi.com/new/doc/

but you'll probably won't need all that.

I'm actually surprised that both control as well as status display are working. That means that roborock hasn't removed the old non-miot commands but apparently just added those as well?


What you need to do now is override the `onMessage` method of your robot implementation (don't forget the super call) and add handling for `properties_changed ` messages there, which most likely for now should just transform the response into the same object returned by the `get_status` method, so that you can just throw that into `parseAndUpdateState`

Hypfer:
lgtm

PR #675, opened 2021-02-03
initial implementation of ViomiZoneCleaningCapability class

rromrrom:
now user is able to edit zones. Zone cleaning does not works at the moment yet.

PR #705, opened 2021-02-20
Viomi go to location capability

rromrrom:
Tested on ViomiV8, works good.
Spot should be created in Zones tab.


lib/robots/viomi/capabilities/ViomiGoToLocationCapability.js
```diff
@@ -0,0 +1,18 @@
...
+class ViomiGoToLocationCapability extends GoToLocationCapability {
+    /**
+     * @param {import("../../../entities/core/ValetudoGoToLocation")} valetudoGoToLocation
+     * @returns {Promise<void>}
+     */
+    async goTo(valetudoGoToLocation) {
+        const point = ViomiMapParser.positionToViomi(valetudoGoToLocation.coordinates.x, valetudoGoToLocation.coordinates.y);
+        Logger.trace("location to go: ", point.x, point.y);
+        await this.robot.sendCommand("set_uploadmap", [0]);
+        await this.robot.sendCommand("set_pointclean", [1, point.x, point.y], {});
```
Hypfer, line 14: Will this actually just move to the location or clean an area around it?

rromrrom: It will go and clean square around 1.5x1.5 meter.

Hypfer: In that case, it's the wrong capability, since GoTo should actually just go to a spot and idle there.

Instead, this would be spot cleanup.

I'm still not convinced that valetudo should implement that feature, because it seems to be just a worse version of the Zoned Cleanup with a predefined zone size.

Thoughts?

There's also #655 for the same thing on roborock

aHVzY2g: I would like to have spot cleanup included, I use it a lot.

rromrrom: If xiaomi/viomi has two separate functions "pointclean" and "zoneclean", then it should be different capabilities in valetudo.

Hypfer: @aHVzY2g But is it any different from drawing a temporary zone around the spot you want to clean?

aHVzY2g: I have to say no, but spot cleaning seems to work more reliable on the ViomiV7. If I draw a temporary zone the robot faces more than often some navigation issues.

lib/robots/viomi/capabilities/ViomiZoneCleaningCapability.js
```diff
@@ -0,0 +1,58 @@
+const ZoneCleaningCapability = require("../../../core/capabilities/ZoneCleaningCapability");
```
rumpeltux, line 1: This file is already part of #675 IIRC, can you remove it from this PR?

lib/robots/viomi/capabilities/ViomiGoToLocationCapability.js
rumpeltux: +1 spot cleaning without having to draw a zone is super useful.

Maybe also worth noting that Viomi also doesn’t have a go-to-xyz-and-do-nothing command (except maybe for go-to-charging-station).

Hypfer: @rumpeltux Yeah but that doesn't necessarily require native support.
The UI could also provide an interface that works like that but actually draws the Zone for the Zoned Cleanup around the spot you're pointing at.

Depau: > @aHVzY2g But is it any different from drawing a temporary zone around the spot you want to clean?

Yes there is.

If you draw an area and there's an obstacle in the middle it will bash on it for a while, as if it were going to jump out of existence if it bashes on it enough times.
If you do a "point clean" it will clean a circle centered on that point (as opposed to a square) and it won't bother too much about any obstacles.

It's most useful to clean a spot close to a wall, if your spot area covers the wall, with the "draw an area" mode it will take a lot longer than the "point clean" mode.

lib/robots/viomi/capabilities/ViomiZoneCleaningCapability.js
tomaszpieczykolan: I think rebasing would get rid of this change from this MR. No need to manually delete those files, and here it would actually create conflicts.

rromrrom: yes, rebase should clean this

Depau:
For what it's worth for this pull request, since GoToLocation has a different meaning for other vacuums, I think it would be best to add a new, very similar `GoCleanLocation` capability and leave `GoToLocation` unimplemented for Viomi.

Other robots could implement this functionality by cleaning a square centered around the selected location.

What do you guys think?

I also have a question:

https://github.com/Hypfer/Valetudo/pull/705/files#diff-4ea56a5209254bf7a29f97efeb15f0e1505ea2fd3ab0d713a4dc6832deff171fR13

```js
        await this.robot.sendCommand("set_uploadmap", [0]);
```

What does this do in this context?

tomaszpieczykolan:
> For what it's worth for this pull request, since `GoToLocation` has a different meaning for other vacuums, I think it would be best to add a new, very similar `GoCleanLocation` capability and leave `GoToLocation` unimplemented for Viomi.
> 
> Other robots could implement this functionality by cleaning a square centered around the selected location.

@Depau I fully agree that it should be a different capability.

Depau (changes requested):
Well this doesn't fix it, you should create a brand-new core capability to handle this case. Renaming it while still inheriting from GoTo only creates confusion.

TimP4w:
What about sending the robot to the location with `set_pointclean` and then stopping the cleaning? This would emulate the desired behaviour and, as a plus, Valetudo would have more capabilities than the official software

Hypfer:
We won't be able to get that to work reliably

Issue #722, opened 2021-02-28
Vacuum Error when conventional (i.e: not water-capable) dust bin is used in Viomi V8


PaskoMoto:
My vacuum only works when the water tank is used. With the conventional dust deposit it fails with an error.

## To Reproduce

1. Fit the conventional dust-only box in the vacuum.
2. The tank is recognized as the voice message says.
3. Click 'Start' in the main web page.
4. The message "Water tank is not installed" is displayed in the main web page. Simultaneously the voice message saying "Make sure the water tank is installed" is reproduced.

### Screenshots

![Captura de pantalla de 2021-02-28 13-02-17](https://user-images.githubusercontent.com/19311679/109417837-8ada1b80-79c5-11eb-8a53-c24125f4519e.png)

### Vacuum Model
Viomi V8 

### Valetudo Version
2021.2.0

## Expected behavior

The vacuum shoud start cleaning normally with the dust-only deposit.

## Additional context
I have recently upgraded from version 0.61. I have noticed several new features, but this specific use-case seems broken to me.
This is the message logged whenever I click the 'Start' button in the web page.
`2021-02-28T11:03:19.931Z] [INFO] Unknown cloud message received: {"method":"event.fault_happen","params":[521]}`

Additional info:
```
device_config.ini:software_version=X1.3.17.55.27R
device_config.ini:hardware_version=1.0.3
device_config.ini:has_rtc_hardware=0

```
Best regards and thanks for this amazing software. ;)

Depau:
I can't reproduce the issue.

The cleaning-type logic is the following:

- if mop is installed
  - if bin has both water tank and dust bin
    - perform vacuum + mop
  - otherwise just mop
- otherwise just vacuum

So I think you're trying to vacuum with the mop installed. Remove it for vacuuming.

Depau:
I tweaked a little the behavior by showing an error message when trying to start cleaning with this accessory configuration.

PaskoMoto:
Ah. Ok.
Didn't know that. I removed the mop and it started to work again with the dust-only deposit,
Many thanks and best regards.

PR #627, opened 2020-11-15
Additional capabilities

Depau:
Draft, as discussed in #584. Blocks #623.

Caps descriptions in https://github.com/Hypfer/Valetudo/pull/584#issuecomment-727614738, progress below:

- [x] `RawCommandsCapability`
- [x] `CarpetTurboModeControlCapability`
- [x] `LEDControlCapability`
- [x] `ManualControlCapability`
- [x] `SensorCalibrationCapability`
- [x] `SpeakerVolumeControlCapability` (was `VolumeControlCapability`, I renamed it to avoid confusion with volume as in m³)
  - [x] Linux ALSA generic implementation
- [ ] `SystemTimeControlCapability`
  - [ ] Linux generic implementation
- [ ] `DoNotDisturbHoursControlCapability`

Optionally also:

- [ ] `CustomVoicePackManagementCapability`
- [ ] `VoicePackManagementCapability`

PR #723, opened 2021-02-28
Viomi fixes and new capabilities

Depau:
~~Work in progress.~~ Ready for review

I will try to find all issues and add more details on the changes.

The plan is to also add (~~maybe in a different PR~~ I eventually implemented them here) the capabilities I planned on implementing (#627)

Changes:

- Fix minor race condition in dummycloud that causes the "Dummycloud is spoofing..." message to not be printed in some cases.
- Add speaker volume control and speaker test capabilities
- Clarify cleaning mode decision to avoid confusion such as #722
- Add carpet turbo mode
- Remove old intensity-based carpet turbo mode
- Add sensor calibration (device accelerometer leveling)
- Fix a bunch of issues with base manual control **base capability** (including removing duration parameter since one would actually need to stop it manually to perform it)
- Implement manual control for viomi
- Add do not disturb
- Removed old `utils/wireless.js` and reimplemented both Viomi **and Dreame** to use the linux wi-fi capability


XobrekX:
I've found another issue on my viomi robot:
After I cleaned some zone(s) and the robot is back at the station, I can't trigger another zone cleaning. I have to trigger a full cleaning. Otherwise the robot won't start the zone cleanup and stays docked in the station. Maybe you wanna have a look there as well :)

Depau:
> I've found another issue on my viomi robot:
> After I cleaned some zone(s) and the robot is back at the station, I can't trigger another zone cleaning. I have to trigger a full cleaning. Otherwise the robot won't start the zone cleanup and stays docked in the station. Maybe you wanna have a look there as well :)

I'm not the author of the zoned cleaning feature for viomi, I don't know how it works and my neighbors won't be very happy if I test it right now, since it's night time.

Can you open an issue and tag me? I'll give it a look when I can.

lib/entities/state/attributes/CarpetTurboStateAttribute.js
```diff
@@ -0,0 +1,32 @@
+const StateAttribute = require("./StateAttribute");
```
Hypfer, line 1: Since this is just a setting, it should not be a state attribute but rather a simple entity in `core/entities`

lib/core/capabilities/ManualControlCapability.js
```diff
@@ -25,10 +25,9 @@ class ManualControlCapability extends Capability {
     /**
      * @abstract
      * @param {string} action
-     * @param {number} [duration] - duration in milliseconds
      * @returns {Promise<void>}
      */
-    async manualControl(action, duration) {
+    async manualControl(action) {
```
Hypfer, line 30: > Fix a bunch of issues with base manual control base capability (including removing duration parameter since one would actually need to stop it manually to perform it)

Please explain further

Hypfer (changes requested):

lib/core/capabilities/ManualControlCapability.js
Depau: I had added the parameter because I inspected python-miio, I saw it supports the duration parameter so I assumed it's an actual feature.

It turns out that this is not the case, python-miio actually sleeps that amount of time and then sends a stop command.
Implementing the same in valetudo doesn't seem ideal.

lib/entities/state/attributes/CarpetTurboStateAttribute.js
Depau: This is just a hack to make it stick around since Viomi does not have a command to retrieve the setting, by adding it to the status it can at least be remembered as long as Valetudo is running.

I know it's not ideal, I'm not sure if there's a better way to do it. I'm also not sure what you have in mind with the entities.

I can remove it completely though and always return an error, if you think it's better.

Depau:
I'm realizing this PR has quickly become quite messy, let me know if you can handle it for review or if you'd rather have me split it into multiple smaller ones.

Hypfer:
> if you'd rather have me split it into multiple smaller ones.

That would be much appreciated

PR #725, opened 2021-03-01
Add voice pack managemenet capability

Depau:
## Type of change
  
Type B:
- [x] New capability

# Description (Type B)
Feature discussion thread: well...

As mentioned in Telegram, I don't think this way of implementing it makes it particularly complex (though slightly less user friendly).

I designed this as a single capability because, at least for the Viomi, separating the "apply local voice pack" and "download new voice pack" functionalities are very similar and creating two capabilities would increase duplication by a lot.

As far as I can tell all vacuums do not accept an upload, but they accept a URL from which they can download the pack. So I exposed it as such, uploading the pack to valetudo and having it serve it back to the vacuum software is quite messy and complex.

Functionality covers:

- Retrieving locally-available voice packs + voice packs that the vacuum or capability can retrieve autonomously
- Applying such voice packs
- Retrieving current voice pack language
- Optionally downloading a custom voice pack from URL
- Retrieving operation progress (which can be checked for both the download of a custom pack and for the application of an available voice pack)

The reason why I did not split this in two is that on Viomi the process of applying a local or remote voice pack is exactly the same.

### DIFF
--- a/lib/core/capabilities/VoicePackManagementCapability.js
+++ b/lib/core/capabilities/VoicePackManagementCapability.js
@@ -1,0 +1,88 @@
+const NotImplementedError = require("../NotImplementedError");
+const Capability = require("./Capability");
+
+class VoicePackManagementCapability extends Capability {
+    /**
+     * This method must return voice packs that are available on the device without requiring any downloads, or voice
+     * packs that the vacuum can download and apply without additional user input.
+     * For example, the Viomi has English and Chinese pre-installed, other voice packs require to be downloaded.
+     * Therefore it should only return English and Chinese at any time, even if a downloaded voice pack is available,
+     * Other vacuums which can talk to the cloud server without additional intervention, or whose capability can perform
+     * the same, may list the available voice packs here.
+     *
+     * @abstract
+     * @returns {Promise<string[]>}
+     */
+    async getAvailableStockVoicePacks() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This method can be used to enable a local, pre-installed stock voice pack, or instruct the vacuum to download one
+     * of its stock voice packs.
+     *
+     * @abstract
+     * @returns {Promise<void>}
+     */
+    async enableStockVoicePack() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * Returns the current applied voice pack language.
+     *
+     * @abstract
+     * @returns {Promise<void>}
+     */
+    async getCurrentVoiceLanguage() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This method should return whether the `downloadCustomVoicePack` functionality is supported by the vacuum.
+     *
+     * @abstract
+     * @returns {Promise<boolean>}
+     */
+    async canDownloadCustomVoicePack() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This method should instruct the vacuum to download a voice pack from `presignedUrl`.
+     * The actual specifications of what exactly is hosted behind presignedUrl depend on the specific vacuum model..
+     *
+     * @abstract
+     * @param {string} language
+     * @param {string} presignedUrl
+     * @returns {Promise<void>}
+     */
+    async downloadCustomVoicePack(language, presignedUrl) {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This method should return the progress of the current voice pack operation.
+     * This could describe one of the following:
+     * - Progress of custom or stock (but not available locally) voice pack download
+     * - Progress of unpacking of in-device voice pack
+     * The progress should be returned as a percentage, null if the operation has completed or no operation in progress.
+     *
+     * @abstract
+     * @returns {Promise<[number]>}
+     */
+    async getVoicePackOperationProgress() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * @returns {string}
+     */
+    getType() {
+        return VoicePackManagementCapability.TYPE;
+    }
+}
+
+VoicePackManagementCapability.TYPE = "VoicePackManagementCapability";
+
+module.exports = VoicePackManagementCapability;
--- a/lib/core/capabilities/VoicePackManagementCapability.js
+++ b/lib/core/capabilities/VoicePackManagementCapability.js
@@ -31,7 +31,7 @@ class VoicePackManagementCapability extends Capability {
      * Returns the current applied voice pack language.
      *
      * @abstract
-     * @returns {Promise<void>}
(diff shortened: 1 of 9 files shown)


Hypfer:
> So I exposed it as such, uploading the pack to valetudo and having it serve it back to the vacuum software is quite messy and complex.

I like that idea.

Can we ditch the local voicepack part and just allow the download and install of a remote URL though?
At least for roborock, there doesn't seem to be this concept of a local cache and I'd say that even if there were one, it wouldn't hurt to redownload every time.

So basically just have

- downloadCustomVoicePack
and
- getVoicePackOperationProgress

Depau:
> > So I exposed it as such, uploading the pack to valetudo and having it serve it back to the vacuum software is quite messy and complex.
> 
> I like that idea.
> 
> Can we ditch the local voicepack part and just allow the download and install of a remote URL though?
> At least for roborock, there doesn't seem to be this concept of a local cache and I'd say that even if there were one, it wouldn't hurt to redownload every time.
> 
> So basically just have
> 
>     * downloadCustomVoicePack
>       and
> 
>     * getVoicePackOperationProgress

Viomi doesn't cache either, it simply has 2 voice packs in the ROM. It only optionally downloads one voice pack at a time and it deletes it when you switch to a new one, including the stock ones in ROM.

I would not remove it because:

- It would make it a lot more troublesome to set one of the stock voice packs, since you'd have to pack them yourself and upload them, which also means they take additional space on a device that already has super tiny flash storage
- It opens up for future implementations that are able to transparently generate URLs to download the official packs from the vendor's cloud

Depau:
Okay, the base capability + router are implemented, I will go ahead and add an implementation for Viomi.

Remarks:

- I kept the built-in voice pack management methods for the reasons detailed above. I also noticed viomi won't let you set english/chinese from a downloaded URL anyway.
- The download voice pack method now takes only presignedUrl as mandatory. Hash and language can optionally be provided. The capability will have to decide what to do if they are required and missing. As discussed in telegram, viomi won't let you set an invalid language code.

<details>
<summary>Viomi language codes</summary>

image

</details>

- I provided a default implementation for the following methods:
  - `getAvailableStockVoicePacks()`: empty list
  - `enableStockVoicePack()`: `Error("Not supported")`
  - `canDownloadCustomVoicePack()`: `false`
  - `downloadCustomVoicePack()`: `Error("Not supported")`
  - `getVoicePackOperationProgress()`: `null` (== no operation in progress)

The REST API is defined as follows:

- `GET /` => `{"currentLanguage": "xx", "operationProgress": int|null}`
- `GET /stockPack` => `[...]` available built-in voice packs
- `PUT /stockPack` body `{"action": "enable", "language": "xx"}`
- `GET /customPack` => `{"supported": bool}`
- `PUT /customPack` body `{"action": "download", "presignedUrl": "http://...", "language": "optional", "hash": "optional"}`

Hypfer:
> which also means they take additional space on a device that already has super tiny flash storage

If a user wants to revert to stock, he might as well just enable the raw command capability and use that.
There's no such thing as baked-in voicepacks on other robots so it doesn't make sense to build that into the capability.

> It opens up for future implementations that are able to transparently generate URLs to download the official packs from the vendor's cloud

No. Under absolutely no circumstance will Valetudo ever talk to the vendor's cloud.


> Error("Not supported")

The whole point of capabilities is that they're implemented fully so that the UI/REST Interface can rely on all features being there if the capability is there. Therefore, this doesn't make sense

Depau:
Alright.

Thinking about it, since no checks are performed against the hash and URL, one could simply provide "local" for both the URL and hash to set a stock voice.

I'll remove that part.

Depau:
New JS API:

- `getCurrentVoiceLanguage()` abstract
- `downloadVoicePack(options)` abstract
- `getVoicePackOperationProgress()` default `null`

New REST API:

- `GET /` => `{"currentLanguage": "xx", "operationProgress": int|null}`
- `PUT /` body `{"action": "download", "presignedUrl": "http://...", "language": "optional", "hash": "optional"}`

On Viomi one can still send `{"action": "download", "presignedUrl": "local", "hash": "local", "language": "en/zh"}` to enable the stock packs.

lib/core/capabilities/VoicePackManagementCapability.js
```diff
@@ -0,0 +1,51 @@
...
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This method should return the progress of the current voice pack operation.
+     * The progress should be returned as a percentage, null if the operation has completed or no operation in progress.
+     *
+     * @returns {Promise<[number]>}
+     */
+    async getVoicePackOperationProgress() {
```
Hypfer, line 37: Since `null` is something best avoided, I'd propose having a `ValetudoVoicePackOperationStatus` entity in `entities/core` with a `type` property that can be

- Idle
- Downloading
- Installing

and a `progress` property that will be the progress as a number in percent 0-100.

I guess idle should be 100% progress?

Hypfer (changes requested):

lib/core/capabilities/VoicePackManagementCapability.js
Depau: Addressed in cc48ab3.

lib/core/capabilities/VoicePackManagementCapability.js
```diff
@@ -0,0 +1,50 @@
...
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This method should return the status of the current voice pack operation, if one is ongoing.
+     *
+     * @returns {Promise<import("../../entities/core/ValetudoVoicePackOperationStatus")>}
+     */
+    async getVoicePackOperationStatus() {
+        return null;
```
Hypfer, line 37: `throw new NotImplementedError();`

lib/robots/viomi/capabilities/ViomiVoicePackManagementCapability.js
```diff
@@ -0,0 +1,71 @@
...
+        return res.curVoice;
+    }
+
+    /**
+     * This method should instruct the vacuum to download a voice pack from `presignedUrl`.
+     * The actual specifications of what exactly is hosted behind presignedUrl depend on the specific vacuum model.
+     * The same goes for the hash, the user should provide a hash or signature as expected by the vacuum.
+     *
+     * @param {object} options
+     * @param {string} options.presignedUrl
```
Hypfer, line 24: I'd go with just `url` here. Nothing presigned necessarily

Hypfer (commented):
Mostly bikeshedding at this point 😁

Depau:
All done :)

PR #732, opened 2021-03-03
viomi: Ensure timezone is UTC

Depau:
As discussed #728

Ensures Viomi is set to UTC after cloud is connected.

Question: should I also adjust DND and schedules to the new timezone? The vacuum doesn't do it on its own, and I can bet we'll see issues "valetudo messed up my schedules".

I decided to keep it simple for now.

### DIFF
--- a/lib/robots/viomi/ViomiValetudoRobot.js
+++ b/lib/robots/viomi/ViomiValetudoRobot.js
@@ -85,6 +85,21 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
         this.tokenFilePath = ViomiValetudoRobot.TOKEN_FILE_PATH;
     }
 
+    onCloudConnected() {
+        super.onCloudConnected();
+
+        this.sendCommand("get_prop", ["timezone"]).then((res) => {
+            if (res.length > 0) {
+                const timezone = res[0];
+                if (timezone !== 0) {
+                    // Set timezone to UTC
+                    this.sendCommand("set_timezone", [0]).then(_ => {
+                    });
+                }
+            }
+        });
+    }
+
     onMessage(msg) {
         switch (msg.method) {
             case "_sync.gen_tmp_presigned_url":


Hypfer:
We may be able to achieve this outside of Valetudo.
It might be interesting to see what the set timezone command actually does on viomi.

On dreame, it just replaces the /etc/localtime symlink with the one of the selected timezone

Therefore, the script that runs valetudo currently contains this:

```
if [ ! "$(readlink /data/config/system/localtime)" -ef "/usr/share/zoneinfo/UTC" ]; then
        rm /data/config/system/localtime
        ln -s /usr/share/zoneinfo/UTC /data/config/system/localtime
fi
```


Maybe we could adapt something similar here?

Depau:
On viomi, `/etc/localtime` is a broken symlink. It actually always keeps the system timezone set to `Asia/Shanghai`, but the time is consistent with the timezone you set from miIO.

Since it is based on OpenWRT, timezone configuration is performed via UCI and stored in `/etc/config/system`, which contains the following content:

```
config system
	option hostname	TinaLinux
	option timezone	Asia/Shanghai
	option timezone CST-8

config timeserver ntp
	list server ntp5.aliyun.com
	list server 0.openwrt.pool.ntp.org
	list server 0.cn.pool.ntp.org
	option enable 1
	option enable_server 0
```

And I notice it uses Chinese servers for NTP, which not only is bad due to the great firewall slowing things down, but there are also the usual "located in China" concerns. How about on Viomi we:

- If embedded
  - Set the timezone first via UCI
  - Replace all NTP servers with `pool.ntp.org` ones
- Finally set the the timezone via miio, regardless of `embedded`, to make sure `RobotApp` is aware of any changes and avoid inconsistencies.

I'm firing up ghidra to see how exactly `RobotApp` handles the timezone.

Hypfer:
Thank your for researching that.
Definitely something to come back to when we start building viomi images using the dustbuilder which will hopefully be very soon.

forast3r:
Hi, is there a fix or some setting that has to be done manually on viomi to get rid of this error message? https://github.com/Hypfer/Valetudo/issues/799

PR #736, opened 2021-03-04
viomi: Fix timezone adjustment race condition

Depau:
ehm as you merged #732 I was adding this. Hang on a second while I also test it to ensure it works, I tested setting the configuration manually and it works.

### DIFF
--- a/lib/robots/viomi/ViomiValetudoRobot.js
+++ b/lib/robots/viomi/ViomiValetudoRobot.js
@@ -92,16 +92,19 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
     onCloudConnected() {
         super.onCloudConnected();
 
-        this.sendCommand("get_prop", ["timezone"]).then((res) => {
-            if (res.length > 0) {
-                const timezone = res[0];
-                if (timezone !== 0) {
-                    // Set timezone to UTC
-                    this.sendCommand("set_timezone", [0]).then(_ => {
-                    });
+        setTimeout(() => {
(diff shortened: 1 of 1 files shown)


Hypfer:
Ah! No I was thinking about doing that in the script/service that starts valetudo. Not valetudo itself

Depau:
Oh I see. In any case I just realized a delay is required after the cloud is connected, I'm gonna add it to the old code.

This is the required sequence, for future reference:

```js
            await spawn("/sbin/uci", ["set", "system.@system[0].zonename=UTC"]);
            await spawn("/sbin/uci", ["set", "system.@system[0].timezone=UTC"]);
            await spawn("/sbin/uci", ["-q", "delete", "system.ntp.server"]);
            await spawn("/sbin/uci", ["add_list", "system.ntp.server=0.pool.ntp.org"]);
            await spawn("/sbin/uci", ["add_list", "system.ntp.server=1.pool.ntp.org"]);
            await spawn("/sbin/uci", ["add_list", "system.ntp.server=2.pool.ntp.org"]);
            await spawn("/sbin/uci", ["add_list", "system.ntp.server=3.pool.ntp.org"]);
            await spawn("/sbin/uci", ["commit", "system"]);
            await spawn("/etc/init.d/sysntpd", ["restart"]);
```

Let me just clear this branch and add back only the timeout

Depau:
Here you go

Depau:
What's this magic that renames the commits? I like it, I want it on all my projects! Though it may break signatures.

Hypfer:
image

PR #739, opened 2021-03-04
Carpet Turbo: project-wide minor changes + viomi implementation

Depau:
Split from #723 

This PR:

- Adds carpet turbo for Viomi. ~~The setting persistence hack is still present: https://github.com/Hypfer/Valetudo/pull/723#discussion_r586827268~~ Carpet mode is now read from the config file if possible, and cached into the Robot class instance.
- Remove old IntensityCapability-based carpet turbo capability
- Handles error messages in the capability router so we can show errors to the user.


lib/entities/state/attributes/CarpetTurboStateAttribute.js
```diff
@@ -0,0 +1,32 @@
+const StateAttribute = require("./StateAttribute");
```
Hypfer, line 1: Quoting from the previous PR:
> Since this is just a setting, it should not be a state attribute but rather a simple entity in `core/entities`


And to answer to your response:

>I know it's not ideal, I'm not sure if there's a better way to do it. 

Due to the magic of javascript, you can simply cache it in a dynamic property of the Viomi Vacuum implementation.
e.g. `this.carpetModeState = new ValetudoCarpetModeState(....)` boom it's there.

I don't want it in the state, because it isn't state. It's just a configuration

>I'm also not sure what you have in mind with the entities.

Entities should be used for things that aren't primitives and also aren't part of the state of the robot.
e.g. Settings such as the `ValetudoDNDConfiguration` which is just a bunch of primitives that require each other to make any sense wrapped into a single thing




And you can scratch some of the above since carpet mode is just a boolean. I'll just keep it here for reference.

Anyways, since carpet mode will most likely stay a single boolean, we can ditch this wrapper altogether.

lib/robots/viomi/capabilities/ViomiCarpetModeControlCapability.js
```diff
@@ -0,0 +1,47 @@
...
+        // 0: off, 1: medium, 2: turbo
+        await this.robot.sendCommand("set_carpetturbo", [2], {});
+    }
+
+    /**
+     * @abstract
+     * @returns {Promise<void>}
+     */
+    async disable() {
+        this.robot.state.upsertFirstMatchingAttribute(new stateAttrs.CarpetTurboStateAttribute({
```
Hypfer, line 39: As explained above, we can simply replace this with a dynamic property on the viomi vacuum implementation.

Disclaimer:
People coming from statically typed languages will probably cry in terror just from reading this suggestion, however personally, I'd argue that this is the point of using a dynamically-typed language such as JS because it lets you do stuff like this without being forced to refactor everything. IMO it's okay to do that for stuff that is self-contained with no side-effects, as long as other options are evaluated and of course code gets refactored if there are more than just a few hacks like these

Anyways:

This should be something like `this.robot.carpetModeEnabled = true` or maybe if there's more of these `this.robot.viomiHacks.carpetModeEnabled` or something like that

Hypfer (changes requested):

lib/robots/viomi/capabilities/ViomiCarpetModeControlCapability.js
Depau: I get your point about it not being state and I agree, as I said I don't think it's a good solution either. However... I also think that setting it as a dynamic property is even nastier. Anyway, your project, your style decisions :)

lib/entities/state/attributes/index.js
```diff
@@ -8,5 +8,5 @@ module.exports = {
     PersistentMapSettingStateAttribute: require("./PersistentMapSettingStateAttribute"),
     OperationModeStateAttribute: require("./OperationModeStateAttribute"),
     AttachmentStateAttribute: require("./AttachmentStateAttribute"),
-    MovementModeStateAttribute: require("./MovementModeStateAttribute")
+    MovementModeStateAttribute: require("./MovementModeStateAttribute"),
```
Depau, line 11: This is a leftover from amending the commit to remove the state attr, whoops

Depau:
State attribute removed in-place and force-pushed, please review ;)

Depau:
I changed it further: the current setting is always stored in the robot instance with the dynamic property, as you suggested.

However, if `embedded === true` and no previous setting is found in the robot class, it attempts to read the vacuum configuration file to retrieve a value. If everything fails it shows an error message "Carpet turbo state cannot be remembered by Valetudo. If you don't remember what you set it to, set it again".

While at it I also noticed that there is absolutely no difference in the proprietary config files between when carpet turbo to `1` or `2`, so it's a good thing we stuck with the boolean capability.

PR #740, opened 2021-03-05
viomi: Manual control capability

Depau:
Split from #723.

Removes `duration` parameter from manual control capability since its implementation would be quite nasty, discussed here: https://github.com/Hypfer/Valetudo/pull/723#discussion_r586777054

Manual control capability is then added to Viomi.

### DIFF
--- a/lib/core/capabilities/ManualControlCapability.js
+++ b/lib/core/capabilities/ManualControlCapability.js
@@ -26,10 +26,9 @@ class ManualControlCapability extends Capability {
     /**
      * @abstract
      * @param {string} action
-     * @param {number} [duration] - duration in milliseconds
      * @returns {Promise<void>}
      */
-    async manualControl(action, duration) {
+    async manualControl(action) {
         throw new NotImplementedError();
     }
 
--- a/lib/robots/viomi/ViomiCommonAttributes.js
+++ b/lib/robots/viomi/ViomiCommonAttributes.js
@@ -1,4 +1,6 @@
 const stateAttrs = require("../../entities/state/attributes");
+const ValetudoSensor = require("../../entities/core/ValetudoSensor");
+const ValetudoManualControlAction = require("../../entities/core/ValetudoManualControlAction");
 
 // Common Viomi enums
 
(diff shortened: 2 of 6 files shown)


tomaszpieczykolan:
Is this ever gonna get merged? 😅

Hypfer:
@Depau could you rebase this PR?

Hypfer:
To document other vendors:

Roborock manual control looks like this:

https://github.com/Hypfer/Valetudo/blob/6db8b76bddfea636303d1169fa7441ac21a5deca/lib/devices/Roborock.js#L323-L327

with an angle, a velocity and a sequence number which needs to be incremented for every command that is part of a manual control session, which needs to be controlled as well

https://github.com/Hypfer/Valetudo/blob/6db8b76bddfea636303d1169fa7441ac21a5deca/lib/devices/Roborock.js#L315-L321

With the S5 Max Mihome controls looking like this:

image
image


And Dreame 1c manual controls look like this:

image

Hypfer:
Rotation-wise, the roborock app seems to always do -60 and +60° 🤔

How much does Viomi rotate on each rotate command?

Depau:
> How much does Viomi rotate on each rotate command?

Roughly 45°

> could you please rebase this PR?

Done :)

Depau:
Viomi manual control:

Screenshot_20210401-192151

Forward and backward do what you expect. Left and right rotate 45° in place, clockwise and counterclockwise.

Since the other robots behave differently, could it make sense to add a method to retrieve the supported movement modes so that the UI can display them properly? For example, a set of these enum values:

- `FORWARD`
- `BACKWARDS`
- `CLOCKWISE`
- `COUNTERCLOCKWISE`

And if some of the robots rotate + go forward when you press left/right

- `LEFT`
- `RIGHT`

Hypfer:
Good idea!

Sounds like another perfect opportunity for the capability properties

https://github.com/Hypfer/Valetudo/blob/a5324e8ece40709be7168730ef77978c3f971d94/lib/core/capabilities/Capability.js#L24-L33

PR #759, opened 2021-03-20
MockCarpetModeControlCapability

ccoors:
## MockRobot features

Type A:
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Adds the `MockCarpetModeControlCapability`

### DIFF
--- a/lib/robots/mock/MockRobot.js
+++ b/lib/robots/mock/MockRobot.js
@@ -11,6 +11,7 @@ class MockRobot extends ValetudoRobot {
         super(options);
 
         this.registerCapability(new capabilities.MockBasicControlCapability({robot: this}));
+        this.registerCapability(new capabilities.MockCarpetModeControlCapability({robot: this}));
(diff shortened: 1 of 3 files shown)


alexkn:
Why do we need a persistent state?

ccoors:
Well, for instance, the initial idea was that if you want to debug an issue that only occurs eg. when the robot is cleaning, the battery is < 50 % and the carpet detection is off, you only have to get your virtual robot into that state once. When you restart Valetudo the previous state is restored.

Idk if it is useful, it's up for discussion.

alexkn:
For this I would not implement a dynamically saved state, but rather multiple config parameters in the implementationSpecificConfig, to bring the robot into a defined state on startup.

What do you think @Hypfer ?

Maybe you should split the PR, MockCarpetModeControlCapability is ok to merge.

Hypfer:
+1 to that

ccoors:
Yeah, you are right. Removed that second commit.

PR #762, opened 2021-03-22
viomi: ViomiMapSegmentationCapability

jloewe:
## Type of change

# Description

I added the ViomiMapSegmentationCapability using `set_mode_withroom`.

The change was tested on my viomi.v7 and works fine, except for the room with the id 11.
For some reason, the robot starts a full cleanup and uses an empty map when this room is specified. Has anyone an idea why this happens?

### DIFF
--- a/lib/robots/viomi/ViomiValetudoRobot.js
+++ b/lib/robots/viomi/ViomiValetudoRobot.js
@@ -95,6 +95,10 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
             robot: this
         }));
 
+        this.registerCapability(new capabilities.ViomiMapSegmentationCapability({
+            robot: this
+        }));
+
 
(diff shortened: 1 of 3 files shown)


Hypfer:
Huh. Segment 11 does seem to work there:

https://community.home-assistant.io/t/xiaomi-viomi-v1-v2-support/165531/17

Maybe Segment 11 _is_ the whole map in your map? Did you check its size?

jloewe:
> Maybe Segment 11 _is_ the whole map in your map? Did you check its size?

I checked it. It is only one "room".

After I played around with my map for a bit whilst testing https://github.com/Hypfer/Valetudo/pull/766 the id of the segment changed and it works fine now.
Both the new segment and the segment with the id 11 seem to work now.

codingstube:
Thanks. Works great and improve valetudo with viomi robot a lot for me.

PR #764, opened 2021-03-22
UI: Show all consumables

ccoors:
## Show all consumables in UI

Type A:
- [x] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Added because the Mop consumable was not displayed at all. The list is now generated dynamically.

### DIFF
--- a/client/settings-consumables.js
+++ b/client/settings-consumables.js
@@ -1,6 +1,21 @@
 /*global ons */
 import {ApiService} from "./services/api.service.js";
 
+const TYPE_MAPPING = Object.freeze({
+    "brush": "Brush",
+    "filter": "Filter",
+    "sensor": "Sensor cleaning",
+    "mop": "Mop"
+});
+
+const SUBTYPE_MAPPING = Object.freeze({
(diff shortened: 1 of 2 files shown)


client/settings-consumables.js
```diff
@@ -19,12 +19,22 @@ async function handleConsumableResetButton(type, subType) {
     }
 }
 
+function setConsumableData(element, res, type, subType) {
+    const data = res.find(e => e.type === type && e.subType === subType);
+    if (data) {
+        element.innerHTML = (data.remaining.value / 60).toFixed(1) + " hours left";
+    } else {
+        element.innerHTML = "N/A";
```
Hypfer, line 27: I'd suggest hiding the elements by default and only making them visible if there is a value for them

ccoors: Okay, the list is getting generated dynamically now.

image

Hypfer: Even better!

PR #766, opened 2021-03-22
viomi: ViomiMapSegmentEditCapability

jloewe:
## Type of change

# Description

I implemented ViomiMapSegmentEditCapability for viomi vacuums
The implementation was tested on my viomi.v7 and works great.

### DIFF
--- a/lib/entities/map/ValetudoMap.js
+++ b/lib/entities/map/ValetudoMap.js
@@ -31,10 +31,7 @@ class ValetudoMap extends SerializableEntity { //TODO: Current, Historic, Etc.
         this.layers = [];
         this.entities = [];
 
-        this.metaData = {
-            version: 1, //Will probably be incremented some day
-            ...options.metaData,
-        };
+        this.metaData.version = 1; // Will probably be incremented some day
 
         if (Array.isArray(options.layers)) {
             this.addLayers(options.layers);
--- a/lib/entities/map/ValetudoMap.js
+++ b/lib/entities/map/ValetudoMap.js
@@ -31,7 +31,10 @@ class ValetudoMap extends SerializableEntity { //TODO: Current, Historic, Etc.
         this.layers = [];
         this.entities = [];
 
-        this.metaData.version = 1; //Will probably be incremented some day
+        this.metaData = {
+            version: 1, //Will probably be incremented some day
+            ...options.metaData,
+        };
 
         if (Array.isArray(options.layers)) {
             this.addLayers(options.layers);
--- a/lib/robots/viomi/ViomiValetudoRobot.js
+++ b/lib/robots/viomi/ViomiValetudoRobot.js
@@ -9,6 +9,10 @@ const ViomiMapParser = require("./ViomiMapParser");
 const zlib = require("zlib");
 
 const stateAttrs = entities.state.attributes;
+const mapActions = Object.freeze({
+    JOIN_SEGMENT_TYPE: 0,
+    SPLIT_SEGMENT_TYPE: 1,
+});
 
 class ViomiValetudoRobot extends MiioValetudoRobot {
     /**
@@ -93,6 +97,12 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
 
         this.registerCapability(new capabilities.ViomiSpeakerVolumeControlCapability({
             robot: this
+        }));
+
+        this.registerCapability(new capabilities.ViomiMapSegmentEditCapability({
+            robot: this,
+            mapActions: mapActions,
+            lang: "en"
         }));
 
 
@@ -381,6 +391,10 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
             }));
         }
 
+        if (data["cur_mapid"] !== undefined) {
+            this.currentMapId = data["cur_mapid"];
+        }
+
         this.emitStateUpdated();
     }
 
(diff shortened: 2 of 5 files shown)


lib/robots/viomi/capabilities/ViomiMapSegmentEditCapability.js
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * @param {import("../../../entities/core/ValetudoMapSegment")} segmentA
+     * @param {import("../../../entities/core/ValetudoMapSegment")} segmentB
+     * @returns {Promise<void>}
+     */
+    async joinSegments(segmentA, segmentB) {
+        await this.robot.sendCommand("arrange_room", {
+            lang: this.lang,
+            mapId: this.robot.currentMapId,
```
Hypfer, line 45: Will this ever be anything that isn't `1`?

jloewe: The currentMapId? For me it currently has a value that looks something like this: `1610520334`

Hypfer: Oh that's the unix timestamp for `Wed Jan 13 2021 06:45:34 GMT+0000`

That correlates with the first time you've used the robot I'd assume?

jloewe: Yep that should match up

Hypfer: Does it fail if it's not set?

jloewe: Yes, I just tried it and the robot really seems not to like it. The response also takes forever.

lib/robots/viomi/ViomiValetudoRobot.js
```diff
@@ -381,6 +391,10 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
             }));
         }
 
+        if (data["cur_mapid"] !== undefined) {
```
Hypfer, line 394: I'm not a fan of this.
With this, we're at two dynamic viomi properties :/

lib/robots/viomi/capabilities/ViomiMapSegmentEditCapability.js
```diff
@@ -0,0 +1,84 @@
...
+        return `${roundedX}_${roundedY}`;
+    }
+
+    /**
+     * @param {import("../../../entities/core/ValetudoMapSegment")} segmentA
+     * @param {import("../../../entities/core/ValetudoMapSegment")} segmentB
+     * @returns {Promise<void>}
+     */
+    async joinSegments(segmentA, segmentB) {
+        await this.robot.sendCommand("arrange_room", {
```
Hypfer, line 43: If the currentMapId is required for this to work, it should throw if it's not available.

Another approach could be to poll the status before. Not sure what would be the best approach here though

lib/robots/viomi/capabilities/ViomiMapSegmentEditCapability.js
```diff
@@ -0,0 +1,84 @@
...
+     * @param {object} pA
+     * @param {number} pA.x
+     * @param {number} pA.y
+     * @param {object} pB
+     * @param {number} pB.x
+     * @param {number} pB.y
+     * @returns {Promise<void>}
+     */
+    async splitSegment(segment, pA, pB) {
+        await this.robot.sendCommand("arrange_room", {
```
Hypfer, line 66: Same thing applies to split as well

Hypfer (changes requested):

lib/robots/viomi/ViomiValetudoRobot.js
jloewe: For some reason `get_props` with the `cur_mapid` is also slow on my robot. Is the mapId contained in the uploaded map that is parsed by the ViomiMapParser?

jloewe: I think I found it in my map dump. The mapId starts at offset 0x04 and is four byte long.

Hypfer: In that case, you can add that to the metaData field of the ValetudoMap returned by the map parser and access that via the robot

plus this check to fail when the map is missing

`if (this.robot.state.map?.metaData?.defaultMap !== true) {`

jloewe: Done. I had to [modify the ValetudoMap to take the metaData option](https://github.com/Hypfer/Valetudo/pull/766/files#diff-40a906408dc9f223acf527c14e56488a5a1e4d7a26fff75f0585ff97f4ed8a3eL34).

lib/entities/map/ValetudoMap.js
```diff
@@ -31,7 +31,10 @@ class ValetudoMap extends SerializableEntity { //TODO: Current, Historic, Etc.
         this.layers = [];
         this.entities = [];
 
-        this.metaData.version = 1; //Will probably be incremented some day
+        this.metaData = {
```
Hypfer, line 34: Are you sure that this is required?

jloewe: I will check it again, but I'm pretty sure

jloewe: Oh sorry I didn't realize that the SerializableEntity handles the metaData

PR #769, opened 2021-03-22
feat(ui): Toggle button states depending on Robot Capabilities

FrankZZ:
Description: 
This PR entails toggling the button states based on the capabilities reported by the Valetudo backend in:
- Home page
- Settings page

I also encountered some minor cosmetic things that I included in this PR such as:
- Add `dBm` unit to the WiFi signal settings field
- Capitalize `WiFi Status`

### DIFF
--- a/client/settings.js
+++ b/client/settings.js
@@ -46,8 +46,6 @@ async function updateSettingsPage() {
 
         Object.keys(buttonStateMap).forEach((key) => {
             const state = buttonStateMap[key];
-            console.log('element ', key)
-            console.log('show', state);
             const element = document.getElementById(`settings-${key}`);
 
             if (element && !state) {
--- a/client/settings.js
+++ b/client/settings.js
@@ -1,31 +1,5 @@
 /*global ons */
 import {ApiService} from "./services/api.service.js";
-/*
-    BasicControlCapability: require("./BasicControlCapability"),
-    CarpetModeControlCapability: require("./CarpetModeControlCapability"),
-    CombinedVirtualRestrictionsCapability: require("./CombinedVirtualRestrictionsCapability"),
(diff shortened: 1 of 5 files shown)


client/settings.js
```diff
@@ -0,0 +1,37 @@
...
+        const robotCapabilities = await ApiService.getCapabilities() || [];
+        const buttonStateMap = {
+            info: true,
+            timers: robotCapabilities.includes("DoNotDisturbCapability"),
+            "carpet-mode": robotCapabilities.includes("CarpetModeControlCapability"),
+            "cleaning-history": false, // commented out in settings.html?
+            "persistent-data": robotCapabilities.includes("PersistentMapControlCapability"),
+            consumables: robotCapabilities.includes("ConsumableMonitoringCapability"),
+            wifi: robotCapabilities.includes("WifiConfigurationCapability"),
+            mqtt: robotCapabilities.includes("WifiConfigurationCapability"),
```
Hypfer, line 15: This can be true all the time

client/settings.js
```diff
@@ -0,0 +1,37 @@
...
+            token: false, // commented out in settings.html?
+            sound: robotCapabilities.includes("SpeakerVolumeControlCapability"),
+            "access-control": robotCapabilities.includes("WifiConfigurationCapability")
```
Hypfer, line 18: Same for access-control

Hypfer:
Hiding elements by default and only unhiding them if they're actually available would be even better.

Now, I see stuff disappearing when I open the settings page

FrankZZ:
> Hiding elements by default and only unhiding them if they're actually available would be even better.
> 
> Now, I see stuff disappearing when I open the settings page

I've addressed this in #774

PR #775, opened 2021-03-24
Update docs

ccoors:
## Doc updates

Type A:
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Various improvements for the docs:
- Outdated Link to `ValetudoRobotFactory`, where the possible `implementation` values are no longer listed
- Explained how to use the `MockRobot` implementation
- Added missing capabilities to capabilities overview


Hypfer:
>Re-generated "Supported Robots" to reflect current capability changes

Since these are generated from the latest master but are visible to newcomers, the idea was to only do that on release so that people don't get confused by seeing features in the docs which aren't there on their robot.

Versioning the docs could of course also be an option but simply not running this script is a lot easier

ccoors:
Okay, removed that commit.

PR #782, opened 2021-03-26
viomi: Fix basic control functionality

### LINKED ISSUE #731: Second zone cleanup not starting - Viomi V7
## Describe the bug
After I cleaned some zone(s) and the robot is back at the station, I can't trigger another zone cleaning. I have to trigger a full cleaning. Otherwise the robot won't start the zone cleanup and stays docked in the station.

## To Reproduce
1. Go to zones
2. Select some random zones and start cleaning
3. Wait till finished 
4. Try to start another zone cleanup
4. See error - robot not starting

### Vacuum Model
Viomi V7

### Valetudo Version
2021.02.0

## Expected behavior
The secoing zone cleanup should start as well

Depau:
## Type of change

Type A:
- [x] Refactor/Code Cleanup
- [ ] Docs

# Description (Type A)

This PR attempts to fix a bunch of issues previously reported, explicitly or implicitly:

Closes #731 

The new basic control capability implementation now remembers what cleaning command was last sent, since it is required later for pausing, stopping and resuming. We have to remember it ourselves since there doesn't seem to be a way to retrieve it from the vacuum state.

This persistence expands on the previously introduced dynamic robot object property for the carpet turbo mode by replacing it with an `ephemeralState` object. This object now stores the last carpet turbo mode and the last cleaning operation details.

This implies that other capabilities that intend to perform cleaning jobs MUST use the methods provided by this cap instead of sending them themselves, to ensure that the state is remembered and to avoid multiple access to the same values. This can be accomplished using a few new public methods which closely match the vacuum counterparts.

The ZoneCleaning capability will be adapted to this new implementation.

----
### EDIT

The implementation is complete but it's too late to test it, please don't merge as long as it's marked as draft.

### DIFF
--- a/lib/robots/viomi/ViomiCommonAttributes.js
+++ b/lib/robots/viomi/ViomiCommonAttributes.js
@@ -21,7 +21,8 @@ const ViomiBoxType = Object.freeze({
 const ViomiOperation = Object.freeze({
     STOP: 0,
     START: 1,
-    PAUSE: 2
+    PAUSE: 2,
+    PAUSE_RECTANGULAR_ZONE: 3
 });
 
 /** @enum {number} */
@@ -32,10 +33,15 @@ const ViomiArea = Object.freeze({
 
 /** @enum {number} */
 const ViomiMovementMode = Object.freeze({
-    NORMAL_CLEANING: 0,
-    MOP_MOVES: 1,  // back and forth mopping movement (unsure if this has an effect without mop-mode)
-    OUTLINE: 2,  // Only clean the rooms outline.
-    ZONED_CLEAN_OR_MOPPING: 3,
+    NORMAL_CLEANING: 0, // goes in straight lines with vacuum motor on
+    VACUUM_AND_MOP: 1,  // back and forth mopping movement with vacuum motor on
+    OUTLINE: 2,         // only clean the rooms outline
+    MOP_NO_VACUUM: 3,   // same as VACUUM_AND_MOP, but the vacuum motor is turned off
+});
+
+const ViomiZoneCleaningCommand = Object.freeze({
+    STOP: 0,
+    CLEAN_ZONE: 3
 });
 
 const FAN_SPEEDS = Object.freeze({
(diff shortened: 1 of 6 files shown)


Depau:
Tested for correct operation, ready for review.

A few changes to ZoneCleaning have been made since:

- Setting the vacuum to zone cleaning mode without doing `set_uploadmap, [1]` (which the app sends when you start drawing a zone) may or may not start cleaning, and if it does it will still report idle and screw up valetudo's state
- Setting vacuum/mop mode after sending the zone list (as opposed to the opposite) also results in this behavior

So the ZoneCleaning cap now performs, in this order:

- `set_mop` (if needed, reusing code from BasicControl)
- `set_uploadmap, [1]`
- `set_zones`
- `set_mode` (using method in BasicControl)

and this is consistent with what the official app does.

Other than that:

- `ViomiMovementMode` value 3 used to be `ZONED_CLEAN_OR_MOPPING` but that doesn't make sense.
  The confusion seems to be caused by the fact that apparently viomi has 2 pause commands (`2` and `3`) which all cleaning commands appear to accept with the same behavior. However the app seems to always pause with 3 zone cleaning, and with 2 all other commands. So I added a new `ViomiOperation` `PAUSE_RECTANGULAR_ZONE` and I use it when pausing that operation.
  Additionally, the command `set_mode` seems to accept two values for the first value, `0` and `3`. When set to `0` it will stop all operations abruptly (and the app uses it as a "fallback" stop command). When set to `3` it controls zone cleaning and it accepts `0`, `1`, and `3` for stop, start and pause.
  Movement mode `3` seems instead to mean `MOP_NO_VACUUM` - it mops with the vacuum motor completely turned off.
- Status polling timeout has been increased to 3 seconds: Valetudo is very busy during cleaning and map parsing slows down normal operation, occasionally this makes it think the command has timed out.
- Status value meanings have been corrected
- I added some short explanations to cryptic vacuum error messages, and added 2 new ones I discovered
- The `ViomiBasicControlCapability` now stores the last cleaning command in `robot.ephemeralState`. It is then used when parsing the status to enrich the cleaning/paused states with the current operation type, and to properly perform pause, stop and resume
- The Zone and MapSegmentation capabilities have been adapted to use the BasicControl capability to start cleaning operations. This way it gets the chance to ensure the correct vacuum/mopping mode and storing the last cleaning command for later use.

Issue #553, opened 2020-07-05
Roborock: Add "Indicator Light" LED disable feature


Hypfer:
Controlled by miio command `set_led_status` and fetched with `get_led_status`, a value of `0` means that the indicator LEDs will be disabled if the robot is docked and fully charged.

This is represented by roborock feature flag `119`

Official App screenshot:
![image](https://user-images.githubusercontent.com/974410/86536811-d0078580-beea-11ea-8776-8b0e70282a36.png)

ccoors:
Which capability would this be? The `LEDControlCapability` does not seem to be the right one (at least in its current state). Maybe adding flags to it? Or inventing a new `IndicatorLightControlCapability`?

Hypfer:
Indeed the `LEDControlCapability` doesn't really make sense how it is now and can probably be removed.

This feature is more like the simple carpet mode toggle and unsurprisingly can also be found right above that in the official app
![image](https://user-images.githubusercontent.com/974410/112116023-23754d00-8bba-11eb-9785-56a5619f2f2c.png)


We could probably refactor both to a common base class `BooleanSettingsControlCapability`, however given that there is basically no code in all of them, that hierarchy almost feels like a university java project

PR #783, opened 2021-03-27
Viomi: Implement LED control capability

Depau:
## Type of change

Type A:
- [X] Refactor/Code Cleanup
- [ ] Docs
- [X] Capability implementation for existing core capability
 
# Description (Type A)

Implements LED control capability for Viomi.

I also adjusted the constructor types for LEDStateAttribute since there were leftovers from when it was copied from ConsumableStateAttribute, and the LED status value didn't make any sense.

### DIFF
--- a/lib/entities/state/attributes/LEDStateAttribute.js
+++ b/lib/entities/state/attributes/LEDStateAttribute.js
@@ -5,18 +5,14 @@ class LEDStateAttribute extends StateAttribute {
      * @param {object} options
      * @param {LEDStateAttributeType} options.type
      * @param {LEDStateAttributeSubType} [options.subType]
-     * @param {object} [options.metaData]
-     * @param {object} options.remaining
-     * @param {number} options.remaining.value
-     * @param {LEDStateAttributeStatus} options.remaining.unit
+     * @param {object} options.metaData
+     * @param {LEDStateAttributeStatus} options.metaData.status
      */
     constructor(options) {
         super(options);
 
         this.type = options.type;
         this.subType = options.subType ?? LEDStateAttribute.SUB_TYPE.NONE;
-
-        this.remaining = options.remaining;
     }
 }
 
--- a/lib/robots/viomi/ViomiValetudoRobot.js
+++ b/lib/robots/viomi/ViomiValetudoRobot.js
@@ -113,6 +113,9 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
             robot: this
         }));
 
+        this.registerCapability(new capabilities.ViomiLEDControlCapability({
+            robot: this
+        }));
 
         this.state.upsertFirstMatchingAttribute(new stateAttrs.AttachmentStateAttribute({
             type: stateAttrs.AttachmentStateAttribute.TYPE.DUSTBIN,
--- a/lib/robots/viomi/capabilities/ViomiLEDControlCapability.js
+++ b/lib/robots/viomi/capabilities/ViomiLEDControlCapability.js
@@ -1,0 +1,54 @@
+const LEDControlCapability = require("../../../core/capabilities/LEDControlCapability");
+const LEDStateAttribute = require("../../../entities/state/attributes/LEDStateAttribute");
+
+/**
+ * @extends LEDControlCapability
+ */
+class ViomiLEDControlCapability extends LEDControlCapability {
+    /**
+     * This function polls the current LEDs state and stores the attributes in our robotState
+     *
+     * @abstract
+     * @returns {Promise<Array<import("../../../entities/state/attributes/LEDStateAttribute")>>}
+     */
+    async getLEDs() {
+        const LEDs = await this.robot.sendCommand("get_prop", ["light_state"]);
+        if (LEDs.length !== 1) {
+            throw new Error("Unable to retrieve LED state");
+        }
+        return [new LEDStateAttribute({
+            type: LEDStateAttribute.TYPE.STATUS,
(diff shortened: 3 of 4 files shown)


Hypfer:
What does this do on Viomi and can we do something similar for https://github.com/Hypfer/Valetudo/issues/553?

Depau:
> What does this do on Viomi and can we do something similar for #553?

It does the same as you described for roborock.

I'm a big fan of the boolean setting capability.

How about we add it, but instead of subclassing it for every possible boolean setting we make it return a list of settings that can be toggled?

Settings as in an object that has a robot-specific ID - it can be anything the robot impl prefers for identification, a type + category, both enums which can be used by the UI to put it inside the correct category and provide the correct title/description.

- `BooleanSettingsCapability`
  - `getSettings()` ⇒ `Array<BooleanSettingDTO>`
  - `getSetting(id)` ⇒ `BooleanSettingDTO`
  - `setSetting(id, value)`
  - `toggleSetting(id, value)`

- `BooleanSettingDTO`
  - `id: number`
  - `type: enum BooleanSettingType`
  - `category: enum BooleanSettingCategory` // or Type + SubType I guess, to be consistent with other DTOs
  - `value: bool`

 If we add it this way I can also easily add the "outline" and "secondary cleanup" for Viomi.

Hypfer:
While that does sound reasonable, it would bypass the capabilities concept by having a capability that needs to be queried further to know what the robot is capable of.

As of now, the UI can just poll the capabilities of the robot to know everything that there is to know about it.
Like this, it would be required to not only poll the list of capabilities but also this (and maybe more) capabilities itself to know what controls to render.

This could work okay-ish for settings such as this LED feature, since I don't expect anyone to change that regularely or via MQTT, however outline and secondary cleanup may actually be better suited to be their own capability?


Unless there are very good reasons to not do that, I'd like to avoid additional layers as much as possible and therefore just have more capabilities

Depau:
> While that does sound reasonable, it would bypass the capabilities concept by having a capability that needs to be queried further to know what the robot is capable of.

Indeed, I see.

> As of now, the UI can just poll the capabilities of the robot to know everything that there is to know about it.
> Like this, it would be required to not only poll the list of capabilities but also this (and maybe more) capabilities itself to know what controls to render.
> 
> This could work okay-ish for settings such as this LED feature, since I don't expect anyone to change that regularely or via MQTT, however outline and secondary cleanup may actually be better suited to be their own capability?
> 
> Unless there are very good reasons to not do that, I'd like to avoid additional layers as much as possible and therefore just have more capabilities

My reasoning for having one single capability was that, if implemented that way, there would be one single capability to be probed for vacuum-specific settings *in the settings page*. Either to display a "device specific settings" settings section or to inject them into one of the existing pages, whatever is more feasible in the front-end.

This "BooleanSetting" capability can be just a dumpster for highly device-specific yet unimportant toggles, offering one simple common implementation from the backend to the front-end that can handle them together.

This has the (very minor tbh) advantage that at any time we can decide some toggle is actually widespread among robots to deserve its own special capability and REST API. The LED capability could actually be one of these, since I would expect pretty much all robots to have similar functionality.

That being said, let me know how you prefer to go forward ;)

More context on the two toggles I mentioned:

- Secondary cleanup, if I remember correctly, is a command you enable once and then it gets automatically disabled on the vacuum side after it starts the second pass.
- Outline-only mode is a modifier to "start" commands (there's a specific movement mode for "outline-only"), the app stores it locally and then it changes the way the start command is sent.

Hypfer:
Hm. A catch-all for vendor specific quirks sounds like a good idea. We should keep that in mind

As for the outline-only mode, I was thinking that maybe there could be a capability which houses the different driving modes?

Those being

- Regular
- Y-Pattern
- Outline-Only
- ???

UI-wise, that would be a drop-down.

Depau:
> Hm. A catch-all for vendor specific quirks sounds like a good idea. We should keep that in mind
> 
> As for the outline-only mode, I was thinking that maybe there could be a capability which houses the different driving modes?
> 
> Those being
> 
>     * Regular
> 
>     * Y-Pattern
> 
>     * Outline-Only
> 
>     * ???
> 
> 
> UI-wise, that would be a drop-down.

I'll see if it's something you can actually choose for Viomi, I'm suspecting only "outline" can be used for all modes, and it will only actually mop if you set the Y pattern.

I'll try and let you know in the next few days.

forast3r:
> > Hm. A catch-all for vendor specific quirks sounds like a good idea. We should keep that in mind
> > As for the outline-only mode, I was thinking that maybe there could be a capability which houses the different driving modes?
> > Those being
> > ```
> > * Regular
> > 
> > * Y-Pattern
> > 
> > * Outline-Only
> > 
> > * ???
> > ```
> > 
> > 
> > UI-wise, that would be a drop-down.
> 
> I'll see if it's something you can actually choose for Viomi, I'm suspecting only "outline" can be used for all modes, and it will only actually mop if you set the Y pattern.
> 
> I'll try and let you know in the next few days.



> Hm. A catch-all for vendor specific quirks sounds like a good idea. We should keep that in mind
> 
> As for the outline-only mode, I was thinking that maybe there could be a capability which houses the different driving modes?
> 
> Those being
> 
> * Regular
> * Y-Pattern
> * Outline-Only
> * ???
> 
> UI-wise, that would be a drop-down.

I'd add another boolean setting in viomi specific options: 
* Repeat (available in mi home app and miiocli "set_repeat", repeats the cleaning cycle only one time).

Issue #640, opened 2021-01-03
Add 'clean' parameter to MQTT configuration (allow persistent connection)


bkk87:
When MQTT settings are configured, Valetudo will **subscribe** to several topics during start-up, e.g. valetudo/rockrobo/**command**.

In Valetudo, it is possible to specify a 'global' QoS parameter. The QoS parameter is correctly set when subscribing or publishing to a topic.

However, what is currently not possible is to set the 'clean' parameter of mqtt.js for topic subscription.
See: https://github.com/mqttjs/MQTT.js

The current default in Valetudo is `clean = true` which means that all messages are deleted when the robot re-establishes a new connection at the MQTT broker.

Being able to set `clean = false` is especially useful when setting QoS to 1 or 2 because this would allow for **offline** messages. 
Example:
 - due to a flacky Wifi, the robot loses the network connection temporarily 
 - a user/automation trigger sends the command to start cleaning
 - robot comes back online
With `clean = false` the robot would see the previous command and start cleaning.
With `clean = true` (current default) , the command message would be deleted at the broker when the robot comes back online and re-establishes the connection.

Hypfer:
This is not abug

Hypfer:
`With clean = false the robot would see the previous command and start cleaning.`

Would it ever stop executing that previous command or just repeat that every time a connection to the broker is established?

bkk87:
If I read the specs correctly then the robot would just start cleaning once and the message is removed/"processed".

The "clean" parameter is different to the "retain" parameter in mqtt. "retain" will always hold the last message and not delete it.

Also, when using the clean parameter, one can configure in e.g. mosquitto that the session will at some point be removed in case the robot does not connect back to the broker (persistent_client_expiration).

PR #787, opened 2021-03-27
MQTT: Homie-compatible implementation

Depau:
## Type of change

Type B:
- [ ] New capability
- [X] New core feature

# Description (Type B)

This pull request aims to provide extensible Homie MQTT support with a focus to not breaking HAss MQTT autodiscovery.

Recap of Homie advantages (see discussion for details)
- Simpler protocol
- Does not use JSON ⇒ removes a class of potential memory corruption issues when a client is implemented in C/C++ i.e. on MCUs
- Jinja templates aren't used ⇒ Orders of magnitude simpler to implement clients in languages ≠ Python
- Compatible with OpenHAB and Node-RED
- Published data is self-explanatory and it does not required a full client library to retrieve and parse.

Feature discussion thread: #724


Depau:
@Hypfer I wanted to bring your attention to this since it's taking shape, if you have some time to take a look and provide some early feedback (nothing works atm but the basic infrastructure is in place).

----

(This comment ended up being lengthier than expected and now I'm too sleepy to summarize further. For a tl;dr just read the bullet points for the main ideas then give a look to the code. I think it's easier to understand than my explanation)

----

Please note that everything related to HAss is pretty much unimplemented or stubbed. I'm gonna go with Homie first since it is the least flexible protocol between the two. I expect to be able to easily plug HAss on top of the same implementation.

During the implementation I decided to go with a slightly different approach. The main idea is the same, here are some remarks:

- I've added `async-mqtt` since the sync API is quite inconvenient. It's a wrapper on top of the previously used `mqtt` library and they can be used at the same time.
- Since Homie has a 3-layer structure but since these 3 layers have more in common than they have not, I decided to implement them with a common base "handle" class (MqttHandle) which is then subclassed and adapted as needed.
- The general idea is each handle manages its own topic, but it can't subscribe and publish to MQTT freely. It will have to go through a strict API. With this I hope to reduce spaghetti and in general have the same benefits as the switch to the capabilities system. If something doesn't work, the API should be extended to adapt the new use case instead of bypassing it and publishing directly.
- As said earlier, the 3 layers are pretty much equal. With this I mean that all the big differences are on the Valetudo side - on the MQTT side all have to subscribe or publish to the topic they manage, and if homie is enabled also provide the homie attributes:
  - Robot => Homie device
    - (with regard to the capabilities)
      -  Capability => Homie node
      - Piece of info / command provided by capabilities => Homie property
    - (with regard to the robot state)
      - Part of the state (such as map, general status) => Homie node
      - Piece of info in the state => Homie property

The setup/reconfiguration/teardown roughly consists of these steps:

- Valetudo launches MqttController
  - MqttController creates and registers to self a new RobotMqttHandle
    - RobotMqttHandle creates and registers to self one instance of a subclass of NodeMqttHandle for each capability, + some more for the state
      - Each Node creates and registers to self as many PropertyMqttHandle as it needs to do its job (but no subclasses - everything is done by passing async callbacks)
   - *[All of the above will happen **before** we connect to MQTT. This will only create a tree]*
   - Once the handle tree is ready, MqttController will connect to MQTT
   - Once connected, it will enter reconfiguration mode and recursively configure all handles
     - Each handle will configure itself and all the children
     - Configuration consists in publishing homie properties (if homie is enabled) and subscribing to setter topics
     - HAss autoconfig will happen at this stage

Teardown/reconnection/reconfiguration:
- MqttController deconfigures RobotMqttHandle recursively
  - this will deconfigure and unsubscribe everything but it will not destroy the handle tree
- MqttController deletes or replaces with a new one its RobotMqttHandle instance
  - The GC will do the rest


#### Reconfiguration mode

Homie expects the device to set its status to "init" while starting up or changing the tree structure. MqttController has a method that takes a callback. Before the callback is called, "init" is reported as the state, then it will go back to "ready" when done.

Inside the callback the handle tree may be rearranged as needed. All handles that need to change will have to be deconfigured then reconfigured at the end.

I expect this API to be useful not only for reloading the configuration, but also for robots that need to change some enum-type parameters dynamically.

#### Pub/sub implementation

Subscribing to MQTT is simple. Each handle can override its own `getInterestingTopics()` method but I don't expect this to be necessary in practice: handles can be created as settable and if that's the case, the default implementation will return the setter topic. This should work for most cases. During configuration this method is probed and all topics subscribed to.
MqttController will dispatch messages to the handles that subscribed to the topic.

Publishing is not yet implemented but there will be a "publish" MqttController method that will wrap the MQTT library. Handles will call this method and pass `this` to it. The controller will retrieve topic + value and publish them.

Handles (already) have a "refresh" method that performs this operation. It can be called from status events callbacks.

#### (De)serialization

For best compliance with homie I'm manually destringifying all incoming and outgoing values based on the datatype (homie has strict typing requirements), currently with zero considerations about hass. I don't expect there to be problems for most types since pretty much all types can be parsed without special considerations in both JS and Python.

Notable exceptions are colors (but we don't care about those, do we?), date-time, durations. For the last two homie mandates them to follow the ISO 8601 standard. I will have to investigate what's the best way to hand these values to HAss. (as a side note, the latest stable version of the homie spec doesn't support datetime and durations. I decided to implement the develop version of the protocol since Valetudo uses them; we can easily add a quirk later to report them as string, if someone complains)

On the JS side, JS can handle ISO date-times but not durations. For those I wrote a tiny parser-writer library.

In any case, I'm pretty confident that the structure I came up with will be flexible enough to take care of this potential incompatibility.

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,248 @@
...
+     */
+
+    /**
+     * Set the controller in reconfiguration mode, then call the callback cb. Attributes, devices and properties can be
+     * safely reconfigured or changed  while inside the callback.
+     *
+     * @param {reconfigureCb} cb
+     * @return {Promise<void>}
+     */
+    async reconfigure(cb) {
```
Hypfer, line 203: An async method that calls a callback?

lib/mqtt/homie/Unit.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * You are free to add more unit types if you so wish. The ones included, not including m² and m³ which
+ * are here just for convenience, are part of the specification and must be recognized by all
+ * Homie-compatible consumers.
+ * If you're dealing with multiples of these units, please perform a conversion instead of adding a
```
Hypfer, line 5: 🙏

lib/res/default_config.json
```diff
@@ -20,7 +20,7 @@
         "enabled": false,
         "server": "foobar.example",
         "port": 1883,
-        "clientId": "",
+        "clientId": "valetudo",
```
Hypfer, line 23: Note that at least mosquitto (and maybe all mqtt brokers) will disconnect an existing session if a new one with the same clientId appears. That could be problematic in multi-robot setups.

Therefore, this should probably become a prefix/suffix to the identifier to avoid these situations

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,248 @@
...
+    /**
+     * @param {object} options
+     * @param {import("../core/ValetudoRobot")} options.robot
+     * @param {import("../Configuration")} options.config
+     */
+    constructor(options) {
+        this.loadConfig();
+
+        this.client = null;
+        // Workaround to be able to set two Last Will and Testament payloads, one for Homie and one for HAss
```
Hypfer, line 18: So it's either Homie or HA?

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,248 @@
...
+        this.clientKey = mqttConfig.clientKey ?? "";
+        this.qos = mqttConfig.qos ?? 0;
+
+        this.homieEnabled = mqttConfig.homie.enabled;
+
+        this.attributesUpdateInterval = mqttConfig.attributesUpdateInterval ?? 60000;
+        this.provideMapData = mqttConfig.provideMapData !== undefined ? mqttConfig.provideMapData : true;
+
+        this.hassEnabled = mqttConfig.hass.enabled; // TODO actually implement hass
+        //this.homeassistantMapHack = mqttConfig.homeassistant.mapHack !== undefined ? mqttConfig.homeassistant.mapHack : true; // TODO
```
Hypfer, line 99: We can drop the Map Hack option and make it on by default with this rewrite, because non-HA users can simply use Homie then and therefore there's no point in being able to turn that off.

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,248 @@
...
+        this.robotHandle = new RobotMqttHandle({
+            robot: options.robot,
+            controller: this,
+            baseTopic: this.topicPrefix,
+            topicName: this.identifier,
+            friendlyName: this.friendlyName
+        });
+
+        if (this.enabled) {
+            this.connect().then();
```
Hypfer, line 45: Retry logic?

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,248 @@
...
+
+
+        this.registerCapabilityAttributeHandlers();
+    }
+
+    /**
+     * @private
+     * @return {object}
+     */
+    getMqttOptions() {
```
Hypfer, line 109: While you're at it, it would be great if you could also evaluate #640

Hypfer (commented):
I am once again amazed by the amount of work you've put not only into getting it done but also getting it done properly and especially giving a useful overview of what you're doing.

It's a pleasure to work with you. Thanks!

lib/mqtt/MqttController.js
Depau: It's not fully implemented yet but it will have publish topics before and after running the callback.
The callback will also have to publish stuff. Since I need the publication to be "blocking" I will use the async API.

This method will be used as follows:

```javascript
await this.controller.reconfigure(async () => {
  this.deregisterChild(stuff);
  this.registerChild(newChild);
  newChild.configure();
});
```

In other words, similarly to python's `with` statement.

Reconfigure will mark the MQTT status as not ready for consumers, run and wait for the callback which will potentially alter the topic structure and Homie attributes, then set the status back to "ready" or something else if errors occur (I haven't decided yet what to do exactly in this case).

lib/res/default_config.json
Depau: That's a good idea. I'll set it to null by default and, if it's null, generate a random string.  To be fair though the client ID was static earlier too. It was using `identifier` instead of `clientID`, `clientId` was already there but it wasn't being used.

I decided to use `identifier` for the robot topic name instead.

lib/mqtt/MqttController.js
Depau: Lol. I realized Homie and Hass want two completely different LWT. So yeah, you need two connections for that.

I temporarily swept the problem under the carpet (~~turbo mode~~) and I'll deal with it better, but I have two solutions in mind:

- Connect to MQTT twice in the same controller; use only one of the connections and use the other one just for the LWT
- Create two instances of the MQTT controller and have one manage homie, the other one hass

I'm more towards the first option, even though it does sound nastier.

lib/mqtt/MqttController.js
Depau: Alright. In general all the commented hass stuff is just there as a reminder to take care of them later, I haven't seen yet what the hack does.

lib/mqtt/MqttController.js
Depau: Yes, definitely planned

lib/mqtt/MqttController.js
Depau: Will do

Depau:
image

 In the screenshot above:
  - `valetudo` is common to all valetudo instances and it can be anything, Homie doens't care, hass had better not care as well
    - `robot` is the device node for this particular robot
      - `fan_speed_control`, `water_grade_control` are the capabilities you know very well, but with a name that doesn't look off in a MQTT broker
        - `fan_speed`, `water_grade` are the properties controlled by the capability. The specification allows for more but these specific capabilities handle one each so there's only one.

Additional remarks:

- I added hooks into `ContainerEntity` for classes to subscribe to status attribute updates
- Since Homie always expects command topics to loop back the set value into the readable topic, I decided it's way more natural for capabilities that trigger status changes, such as Intensity*, to manage their piece of status on the MQTT side
- I decided to not use the capability class name as the MQTT topic cause I think it's ugly. I'm still in time to change it, should I?

Hypfer:
> I'm still in time to change it, should I?

I'd prefer having the uglier but consistent capability class names instead of introducing another name translation layer

> The CI build is breaking but it's not my code's fault

Huh. Indeed. I'll look into that

Hypfer:
I've found my comment on using the Capability Names:

https://github.com/Hypfer/Valetudo/pull/584#issuecomment-677815648

Depau:
@Hypfer Since I changed the MQTT structure, is it okay if I try to do a migration if the old config is detected?

Anyway, most stuff is ready. The last missing pieces are checking for disconnections once we're up and running and some documentation on how to write new MQTT handles since....

image


I am proud to announce that (in my opinion), I managed to come up with an OOP structure that enforces keeping the strict Homie topic tree and mapping 1:1 to the capability system, while at the same time providing HAss support in a way that will/should make people uncomfortable if they try to support only either technology or try to do "nasty hacks".

This is accomplished by first setting up the homie-style handle-based structure, then attaching a hass component to any handle that "makes sense".

For instance, the component for `vacuum.mqtt` is attached to the main homie Robot handle (the root handle). The handle for the water grade capability is attached to the property handle that manages the intensity.

With this structure you can easily share the capability/attribute status topic between the two: since the hass component is attached to the handle it shares data with, the handle topic can be immediately extracted and shoved into the hass component autoconfig payload.

A custom payload can be provided for a hass component to handle cases such as the vacuum state or `json_attributes`. This payload may or may not contain `HassAnchor`s - global, unique, dynamic references to information or topics already managed by other handles. If they do so, the payload will be refreshed automatically when new data is posted to the Anchor by the handle that owns control of it.

lib/mqtt/MqttController.js
Depau: It turns out that MQTT.js handles this internally. Both cases "broker unavailable on start" and "broker becomes unavailable" are handled, it will just spam the log once per second but I think it's okay.

```
[2021-04-05T01:16:44.496Z] [ERROR] MQTT error: Error: connect ECONNREFUSED 10.69.0.13:1883
[2021-04-05T01:16:45.499Z] [ERROR] MQTT error: Error: connect ECONNREFUSED 10.69.0.13:1883
[2021-04-05T01:16:46.499Z] [ERROR] MQTT error: Error: connect ECONNREFUSED 10.69.0.13:1883
[2021-04-05T01:16:47.502Z] [ERROR] MQTT error: Error: connect ECONNREFUSED 10.69.0.13:1883
[2021-04-05T01:16:48.504Z] [ERROR] MQTT error: Error: connect ECONNREFUSED 10.69.0.13:1883
```

Depau:
This is now ready for review. Since it's a chonky PR and it's mostly additions I would suggest to clone it and read the code from the filesystem. I can't do much splitting since it's literally all MQTT.

You may however want to check the diffs for stuff under `/lib/entities`.

I added a dev documentation page in the very last commit, explains the structure. I tried to do a better job than the comment from a few days ago.

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,595 @@
...
+     * @param {string} topic
+     * @param {string} message
+     * @param {object} [options]
+     * @return {Promise<void>}
+     */
+    async publishHass(topic, message, options) {
+        const err = new Error();
+        Error.captureStackTrace(err, this.publishHass);
+        const lines = err.stack.split("\n");
+        if (lines !== undefined && lines.length > 1 && !lines[1].includes(HassController.name)) {
```
Hypfer, line 580: Lol this level of self-defense

We should keep in mind that creating a stacktrace is a somewhat expensive operation, although that shouldn't really matter here, because it will only be called every few seconds, I think?

lib/mqtt/handles/MqttHandle.js
```diff
@@ -0,0 +1,377 @@
...
+     */
+    constructor(options) {
+        this.parent = options.parent === undefined ? null : options.parent;
+        this.controller = options.controller;
+        this.topicName = options.topicName;
+        this.friendlyName = options.friendlyName;
+        this.dataType = options.datatype;
+        this.format = options.format;
+
+        this.retained = options.retained === undefined ? true : options.retained;
```
Hypfer, line 33: Since we're using node v14 and upwards, we can use the nullish coalescing operator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

```suggestion
        this.retained = options.retained ?? true;
```

This also applies to all other instances of similar ternary usage

lib/mqtt/handles/RobotMqttHandle.js
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * @public
+     * @return {null|MapNodeMqttHandle}
+     */
+    getMapHandle() {
+        return this.mapHandle;
+    }
+}
+
+const CAPABILITY_TYPE_TO_HANDLE_MAPPING = {
```
Hypfer, line 143: Would it make sense to move these two maps into seperate files?

lib/mqtt/homeassistant/ComponentType.js
```diff
@@ -0,0 +1,19 @@
+const ComponentType = Object.freeze({
```
Hypfer, line 1: Maybe add a comment that references the source and date of retrieval of these?

lib/mqtt/MqttController.js
```diff
@@ -0,0 +1,595 @@
...
+     * falling back to a randomly generated one.
+     *
+     * @private
+     * @return {string}
+     */
+    genClientId() {
+        // @ts-ignore - deviceId only exists in MiioValetudoRobot
+        const devideId = this.robot.deviceId;
+        if (devideId === undefined) {
+            return "valetudo_" + Math.random().toString(16).substr(2, 8);
```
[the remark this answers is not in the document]
Depau: Roborock should have a miio device ID, shouldn't it? The random device ID is only used if we're not using `MiioValetudoRobot`, if we have a miio deviceID we hash it and use the first few characters of the hash.

Also, I just noticed the typo. That definitely needs fixing :)

lib/mqtt/MqttController.js
Depau: I just have this feeling that someone at some point is gonna use it to duct-tape some weird behavior for some "special" hass component and that it will slip through review.

But maybe it's just:

> +++ BREAKING NEWS: "Man invents fictional scenario and gets angry about it" +++

I guess it can be dropped :)

lib/mqtt/handles/RobotMqttHandle.js
Depau: Yes, that would probably make it less confusing.

Depau:
One question:

Since I changed enough stuff in the configuration that will make it crash on start if started with an old config, should I implement some sort of automatic migration?

Hypfer:
Ah right yeah. Hm maybe? Depends on how much work it is to do that.

If yes, we should define how long that code shall exist. Just in the next release or maybe two releases?

Depau:
I'd say 2 releases are more than enough.

Actually let me also adjust the the config UI so it controls the new settings, otherwise people on `master` will get crashes when they try to change stuff.

Depau:
Okay, all done.

Let me know what to do with the autogenerated clientId. Do roborocks really have all the same deviceId?

Hypfer:
Ah no the deviceId is unique for each device. I was just thinking about cases where there is no deviceId. Although we should be able to postpone that until a non-miio robot appears.

As for the CI failing: a9d1b14 might fix this

Depau:
Nope :/

package-lock.json
```diff
@@ -1,8 +1,5336 @@
...
+      "version": "2.2.2",
+      "resolved": "https://registry.npmjs.org/picomatch/-/picomatch-2.2.2.tgz",
+      "integrity": "sha512-q0M/9eZHzmr0AulXyPwNfZjtwZ/RBZlbN3K3CErVrk50T2ASYI7Bye0EvekFY3IP1Nt2DHu0re+V2ZHIpMkuWg==",
+      "dev": true,
+      "engines": {
+        "node": ">=8.6"
+      }
+    },
+    "node_modules/pkg": {
+      "resolved": "git+ssh://git@github.com/Hypfer/pkg.git#7c99421c887f6a36981b1fb0f92d89bbb68e8528",
```
Hypfer, line 3702: These git+ssh:// seem to be the culprit as far as I can tell.

git+https:// should probably work

Depau: Build now works :)

Hypfer:
So the previous homeassistant autoconf topic was
`homeassistant/vacuum/valetudo_rockrobo/config`
with the one now being
`homeassistant/vacuum/rockrobo/rockrobo_vacuum/config`

Would it be possible to change that back to the old schema? Because as it is now, I'm getting duplicate vacuum entities when switching to the new version

Depau:
> So the previous homeassistant autoconf topic was
> `homeassistant/vacuum/valetudo_rockrobo/config`
> with the one now being
> `homeassistant/vacuum/rockrobo/rockrobo_vacuum/config`
> 
> Would it be possible to change that back to the old schema? Because as it is now, I'm getting duplicate vacuum entities when switching to the new version

Yeah I don't think there's any problem with that, according to the docs it should accept it both ways.

PR #802, opened 2021-04-01
Feature/UI home segments

### LINKED ISSUE #773: UI Feature: Clean named Segment from Homepage
If the `MapSegmentationCapability` capability is present, we can fetch all Segments from `/api/v2/robot/capabilities/MapSegmentationCapability`, filter for those which are named and build a multi-select capable list of them just like with the ValetudoZonePresets:

image

Jomik:
## Type of change
UI implementation of capability

# Description 
Shows a dialog, similar to Zones, to allow you to select named segments, and clean them.
Implemented using "source-code copy/paste", as refactoring this with all the global variables is a pain.

Fixes #773

### DIFF
--- a/lib/robots/mock/capabilities/MockMapSegmentationCapability.js
+++ b/lib/robots/mock/capabilities/MockMapSegmentationCapability.js
@@ -1,0 +1,32 @@
+const MapSegmentationCapability = require("../../../core/capabilities/MapSegmentationCapability");
+const ValetudoMapSegment = require("../../../entities/core/ValetudoMapSegment");
+
+/**
+ * @extends MapSegmentationCapability<import("../MockRobot")>
+ */
+class MockMapSegmentationCapability extends MapSegmentationCapability {
+    /**
+     * @returns {Promise<Array<import("../../../entities/core/ValetudoMapSegment")>>}
+     */
+    async getSegments() {
+        return [
+            new ValetudoMapSegment({ id: "foo_id", name: "Foo"}),
+            new ValetudoMapSegment({ id: "bar_id", name: "Bar"}),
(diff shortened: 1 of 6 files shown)


lib/robots/mock/capabilities/MockMapSegmentationCapability.js
```diff
@@ -0,0 +1,32 @@
...
+ */
+class MockMapSegmentationCapability extends MapSegmentationCapability {
+    /**
+     * @returns {Promise<Array<import("../../../entities/core/ValetudoMapSegment")>>}
+     */
+    async getSegments() {
+        return [
+            new ValetudoMapSegment({ id: "foo", name: "Foo"}),
+            new ValetudoMapSegment({ id: "boo", name: "Bar"}),
+            new ValetudoMapSegment({ id: "empty" })
```
SBeyeMHP, line 15: Uh? Empty?

SBeyeMHP (changes requested):

lib/robots/mock/capabilities/MockMapSegmentationCapability.js
Hypfer, line 15: Why does the backend return the "empty" label?

Hypfer (changes requested):

lib/robots/mock/capabilities/MockMapSegmentationCapability.js
Jomik: I guess I could have written something else. It is just an id for the segment with an empty/undefined name.

Jomik: I changed the IDs to avoid confusion.

PR #820, opened 2021-04-10
Viomi minor bugfixes

Depau:
- Timezone adjustment is moved to the status parsing so that we don't get any race conditions on cloud connection
- Fix a regression in map parsing which causes it to loop forever if the end of the not-well-understood room data structure cannot be found
- ~~Makes map "presigned URL" payload consistent with how the OEM cloud currently sends it~~ moved to #873

### DIFF
--- a/lib/robots/viomi/ViomiMapParser.js
+++ b/lib/robots/viomi/ViomiMapParser.js
@@ -321,7 +321,7 @@ class ViomiMapParser {
 
                     mapContents.path.points[mapContents.path.points.length - 2] -
                     mapContents.path.points[mapContents.path.points.length - 4]
-                ) * 180 / Math.PI) + 90) % 360; //TODO: No idea why
+                ) * 180 / Math.PI) + 270) % 360; //TODO: No idea why
             }
         }
 
--- a/lib/robots/viomi/ViomiMapParser.js
+++ b/lib/robots/viomi/ViomiMapParser.js
@@ -463,10 +463,15 @@ class ViomiMapParser {
             // This simply tries to eat up as many bytes until the start of the known tag value is found. Note that the tag
             // varies from vacuum to vacuum
             let takenBytes;
+            let count = 0;
             do {
                 takenBytes = this.peek(4);
                 this.offset += 1;
-            } while (Buffer.compare(takenBytes, rooms2header.slice(0, 4)));
+                count += 1;
+                if (count >= 300) {
+                    throw new Error("Unable to seek to end of room data");
+                }
+            } while (Buffer.compare(takenBytes, rooms2header.slice(0, 4)) && this.offset < this.buf.length);
 
             this.take(3, "rest of tag");
         }
--- a/lib/robots/viomi/ViomiMapParser.js
+++ b/lib/robots/viomi/ViomiMapParser.js
@@ -201,11 +201,13 @@ class ViomiMapParser {
             this.take(8, "unknown8");
             this.parseRooms();
 
-            this.points = [];
-            try {
-                this.parsePose();
-            } catch (e) {
-                Logger.warn("Unable to parse Pose", e); //TODO
+            if (this.mapId !== 0) {
+                this.points = [];
+                try {
+                    this.parsePose();
+                } catch (e) {
+                    Logger.warn("Unable to parse Pose", e); //TODO
+                }
             }
         }
         this.take(this.buf.length - this.offset, "trailing");
@@ -237,7 +239,7 @@ class ViomiMapParser {
      * @return {number}
      */
     viomiToValetudoAngle(angle) {
-        let result = (-180 -(angle * 180 / Math.PI)) % 360;
+        let result = (-180 - (angle * 180 / Math.PI)) % 360;
         while (result < 0) {
             result += 360;
         }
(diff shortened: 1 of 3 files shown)


lib/robots/viomi/ViomiValetudoRobot.js
```diff
@@ -211,23 +193,38 @@ class ViomiValetudoRobot extends MiioValetudoRobot {
...
-                let mapUploadUrls = [];
-                for (let i = 0; i < 4; i++) {
-                    let url = this.mapUploadUrlPrefix +
-                        "/api/miio/map_upload_handler?" +
-                        process.hrtime().toString().replace(/,/g, "") +
-                        "&suffix=" + key;
-                    mapUploadUrls.push({"url": url, "method": "PUT"});
+                let result = {ok: true};
+
+                let url = this.mapUploadUrlPrefix + "/api/miio/map_upload_handler?" +
```
Hypfer, line 198: Roborock uses the same commands but with a different looking payload it seems?

https://github.com/Hypfer/Valetudo/blob/07a1c7ad78e3593b90e12ba0bb8286ed0c4f58e1/lib/robots/roborock/RoborockValetudoRobot.js#L102-L115

But dreame seems to be the same
https://github.com/Hypfer/Valetudo/blob/07a1c7ad78e3593b90e12ba0bb8286ed0c4f58e1/lib/robots/dreame/DreameValetudoRobot.js#L50-L80

Can we refactor this?

Hypfer: Also, I wonder if roborock would also work with the same structure. Considering that two different vendors use that, it might be standard xiaomi stuff?

Depau: That's possible. This is what I get on viomi with the stock app:

vac → cloud: `{"id":295,"method":"_sync.gen_presigned_url","params":{"suffix":"bin"}}`
cloud → vac: 
```json
{"id":295,"result":{"bin":
{"url":
"http://awsde0.fds.api.xiaomi.com/viomi-device/2021/04/10/####/######.bin?Expires=1618049385000&GalaxyAccessKeyId=#######&Signature=zH0kgTPE1XUQPzR
74Vv2vsmAbtI=",
"obj_name":"2021/04/10//####/######.bin",
"method":"PUT","expires_time":1618049385,"ok":true,
"pwd":"changes every time"},
"ok":true}}`

Comment (Depau) [reply lib/robots/viomi/ViomiValetudoRobot.js:198]: Yeah `_sync.gen_presigned_url` appears to match dreame. I haven't really checked for `gen_tmp_presigned_url` so I left it as it is since it's working. I'll see.

I was noticing some issues where it sometimes doesn't update the map with new obstacle info, I was investigating whether the map URL payload affects it (it doesn't)

Comment (Hypfer) [reply lib/robots/viomi/ViomiValetudoRobot.js:198]: Just checked roborock again on a cloud-provisioned S5 Max

vac → cloud: `{"id":1234,"method":"_sync.batch_gen_room_up_url","params":{"indexes":[12,13,14,15]}}`
cloud → vac: 
```json
{
	"id": 1234,
	"result": [
		"https://awsusor0.fds.api.xiaomi.com/robomap/rubyslite%2F###DID###%2F12?Expires=1618071662000&GalaxyAccessKeyId=#######&Signature=#######",
		"https://awsusor0.fds.api.xiaomi.com/robomap/rubyslite%2F###DID###%2F13?Expires=1618071662000&GalaxyAccessKeyId=#######&Signature=#######",
		"https://awsusor0.fds.api.xiaomi.com/robomap/rubyslite%2F###DID###%2F14?Expires=1618071662000&GalaxyAccessKeyId=#######&Signature=#######",
		"https://awsusor0.fds.api.xiaomi.com/robomap/rubyslite%2F###DID###%2F15?Expires=1618071662000&GalaxyAccessKeyId=#######&Signature=#######"
	]
}
```

Depau: I don't see any escapes.

Anyway, viomi doesn't seem to use this `batch_gen_room_up_url` thing, but I guess we can definitely try to move the map upload to `MiioValetudoRobot` and see what breaks.

lib/robots/viomi/ViomiMapParser.js
```diff
@@ -458,6 +488,11 @@ class ViomiMapParser {
             //Repeated for every room there is
             this.take(unkLength * 2, "room data");
 
+            if (this.mapId === 0) {
+                // Sometimes the map is truncated here, when that happens the second word is zeroes
```
Hypfer, line 492: Could these be incremental map updates or do those still contain a full map image etc?

Depau: Nope, it contains everything except the "extended" room data + room outlines.

For some reason this is sent as soon as the robot touches the dock and in no other circumstance.

Hypfer: Maybe thats when room splitting stuff is recalculated or something like that 🤔

Depau: Indeed I just noticed it wants the same map ID for editing segments. It probably returns map ID set to 0 so that segment editing won't work until it's not done.

Depau:
Since the presigned URL stuff was moved to #873, this can now be reviewed.

PR #840, opened 2021-04-18
docs(mqtt): Autogenerated MQTT docs

Depau:
- [x] ToC
- [x] Indices
- [x] Capability docs
- [x] Status attributes docs
- [x] Links to capability overview pages
- [x] Remove linked SVG and include in local assets
- [x] Implement mock capabilities for stuff that is implemented in MQTT but is currently undocumented due to missing capabilities for mock robot


Hypfer:
Would it be possible to maintain the helpText in a file that isn't part of the actual valetudo build and just join them by className when building the docs?

Otherwise, we might bloat the binary which stuff that doesn't really belong there

Depau:
> Would it be possible to maintain the helpText in a file that isn't part of the actual valetudo build and just join them by className when building the docs?
> 
> Otherwise, we might bloat the binary which stuff that doesn't really belong there

Yes, how about a yaml so we at least get multiline strings for free?

Depau:
Update for the records, we concluded that it's way simpler to just keep the help text in the capabilities in order to uncomfortable lookups in the yaml.

Here's a preview: https://md.depau.eu/9eMyhiO5Sb-G8yjaSjEfDA?view

Depau:
I'd say all done.

The HA page needs adjustments since everything relies on `vacuum.send_command`, maybe this should be used instead? https://www.home-assistant.io/docs/mqtt/service/

Depau:
Changes preview here: https://depau.github.io/Valetudo/

PR #842, opened 2021-04-20
feat(vendor.mqtt): Add MapSegmentationCapabilityMqttHandle

Depau:
## Type of change

Type B:
- [ ] New capability
- [x] New core feature

Add segment cleaning over MQTT. Payload: either `SEGMENT_ID` or `[SEGM1, SEGM2, ... , SEGM#]`


Hypfer:
Could you also add the functionality for Home Assistant?

I think it's currently not possible to do segment cleanups from there using the current master?

Depau:
> Could you also add the functionality for Home Assistant?
> 
> I think it's currently not possible to do segment cleanups from there using the current master?

There's nothing to do for home assistant, there's no component that can do it. People need to use mqtt.publish and publish to the `/set` topic, just like the zone presets.

PR #846, opened 2021-04-21
docs(mqtt): Add better openHAB docs and custom widget

Depau:
## Type of change

Type A:
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Adds docs for adding my custom widget in openHAB.

Note that the page intentionally goes over details already covered by the official OH docs: this is to avoid people setting it up in incompatible ways and set a common standard for the IDs of linked items, since you are free to call them whatever you want.

This is to encourage creating more community custom Valetudo widgets for openHAB that require the same compatible naming and item suffixes, and that are therefore easier to share and maintain.

Preview: https://depau.github.io/Valetudo/pages/integrations/openhab-integration.html

Not included in the docs since it's should be pretty self-explanatory but these are the available config options:

image

image

image


Depau:
Changed to draft since I'm going to add segments retrieval to the back-end first, then use it in the widget.

Though I guess we could have a repo or gist for the widgets, what do you think?

Hypfer:
Great idea!

=> https://github.com/Hypfer/Valetudo-Openhab/

PR #852, opened 2021-04-24
feat(core): Add RobotSpecificToggleCapability

Depau:
## Type of change

Type B:
- [x] New capability
- [ ] New core feature

RFC before I continue to add the rest of what's required.

Thinking about my comment about it on Telegram:

> I'm actually not sure though, because on the MQTT side it would be "nice to have" but it would be kind of hard to sync the topics since they wouldn't be part of the status

I eventually concluded that it's easier to just store it in the status. Front-ends may consider the status value good if it's present, but otherwise they should call the capability to update it (the `getToggle` and `setToggle` methods should update the status accordingly in implementations).

---

Motivations for having this capability:

- It works well for the LEDs toggle
- `set_repeat`:
  > The spinning side brush tends to launch tiny chunks where it's already been, so I usually have it repeat the cleaning to get those too.
  > Also sometimes I send it to mop the bathroom to dry water splashes, one pass usually isn't enough.
  > Right now with Valetudo I still do it, it's just more annoying since I have to wait for it to finish and start it again. Which is okay during lockdown, but in normal conditions I used to have it run while I'm out.

### DIFF
--- a/lib/core/capabilities/RobotSpecificToggleCapability.js
+++ b/lib/core/capabilities/RobotSpecificToggleCapability.js
@@ -1,0 +1,63 @@
+const Capability = require("./Capability");
+const NotImplementedError = require("../NotImplementedError");
+
+/**
+ * @template {import("../ValetudoRobot")} T
+ * @extends Capability<T>
+ */
+class RobotSpecificToggleCapability extends Capability {
+    /**
+     * Returns the toggle types this robot supports
+     *
+     * @abstract
+     * @returns {Array<string>}
+     */
+    getSupportedToggleTypes() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * This function updates and returns the supported robot specific toggles and their values.
+     *
+     * @returns {Promise<Array<import("../../entities/state/attributes/RobotSpecificToggleStateAttribute")>>}
+     */
+    async getToggles() {
+        const result = [];
+        for (const type of this.getSupportedToggleTypes()) {
+            result.push(await this.getToggle({type: type}));
+        }
+        return result;
+    }
+
+    /**
+     * This function updates and returns one supported robot specific toggle and its values
+     *
+     * @abstract
+     * @param {object} options
+     * @param {string} options.type
+     * @return {Promise<import("../../entities/state/attributes/RobotSpecificToggleStateAttribute")>}
+     */
+    async getToggle(options) {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * Sets a robot toggle to a specific value.
+     *
+     * @param {object} options
+     * @param {string} options.type
+     * @param {boolean} options.value
+     * @returns {Promise<void>}
+     */
+    async setToggle(options) {
+        throw new NotImplementedError();
+    }
+
+    getType() {
+        return RobotSpecificToggleCapability.TYPE;
+    }
+}
+
+RobotSpecificToggleCapability.TYPE = "RobotSpecificToggleCapability";
+
+module.exports = RobotSpecificToggleCapability;
--- a/lib/core/capabilities/index.js
+++ b/lib/core/capabilities/index.js
@@ -17,6 +17,7 @@ module.exports = {
     MapSegmentationCapability: require("./MapSegmentationCapability"),
     MapSnapshotCapability: require("./MapSnapshotCapability"),
     PersistentMapControlCapability: require("./PersistentMapControlCapability"),
+    RobotSpecificToggleCapability: require("./RobotSpecificToggleCapability"),
     SensorCalibrationCapability: require("./SensorCalibrationCapability"),
     SpeakerTestCapability: require("./SpeakerTestCapability"),
     SpeakerVolumeControlCapability: require("./SpeakerVolumeControlCapability"),
--- a/lib/entities/state/attributes/RobotSpecificToggleStateAttribute.js
+++ b/lib/entities/state/attributes/RobotSpecificToggleStateAttribute.js
@@ -1,0 +1,32 @@
+const StateAttribute = require("./StateAttribute");
+
+/**
+ * Represents a robot specific toggle.
+ * Having it in the state makes it easier to propagate the changes to MQTT. However, front-ends should use
+ * RobotSpecificToggleCapability.getToggles() to ensure they get up-to-date values.
+ */
+class RobotSpecificToggleStateAttribute extends StateAttribute {
+    /**
+     * @param {object} options
+     * @param {RobotSpecificToggleStateAttributeType} options.type
+     * @param {boolean} options.value
+     * @param {object} [options.metaData]
+     */
+    constructor(options) {
+        super(options);
+
+        this.type = options.type;
+        this.value = options.value;
+    }
+}
+
+/**
+ *  @typedef {string} RobotSpecificToggleStateAttributeType
+ *  @enum {string}
+ */
+RobotSpecificToggleStateAttribute.TYPE = Object.freeze({
+    LEDS: "leds",
+    REPEAT_CLEANING: "repeat_cleaning",
+});
+
+module.exports = RobotSpecificToggleStateAttribute;
--- a/lib/entities/state/attributes/index.js
+++ b/lib/entities/state/attributes/index.js
@@ -8,5 +8,6 @@ module.exports = {
     MovementModeStateAttribute: require("./MovementModeStateAttribute"),
     OperationModeStateAttribute: require("./OperationModeStateAttribute"),
     PersistentMapSettingStateAttribute: require("./PersistentMapSettingStateAttribute"),
+    RobotSpecificToggleStateAttribute: require("./RobotSpecificToggleStateAttribute"),
     StatusStateAttribute: require("./StatusStateAttribute")
 };


Hypfer:
> It works well for the LEDs toggle

But those should be a real capability because that's not vendor specific. There are at least two vendors which do have such a thing.

> set_repeat

Even if it's contained in one specific capability, I don't like having such vendor-specific quirks in the generic abstraction that is Valetudo, because it will sooner or later leave that containment and cause bad things in other parts of the codebase such as

> I eventually concluded that it's easier to just store it in the status.


See? It's already happening

Hypfer:
Especially with Viomi, there need to be very good reasons for why Valetudo should consider adopting something for it, because it's not a real vendor with just a single product that is already EOL based on a white-label thing with a hacked-together miio software stack

Supporthing that beyond what is possible by implementing existing capabilites just doesn't make any sense for Valetudo even though I of course understand that some things might be annoying for Viomi users.
I'm just trying to prevent footguns here

Depau:
> cause bad things in other parts of the codebase such as
>>    I eventually concluded that it's easier to just store it in the status.
>
> See? It's already happening

I get your point here and I agree to some extent. However, I fail to see most of the problems:

- Having multiple capabilities only adds boilerplate for functionality that works exactly the same way - abstracting them into a common parent (as opposed to having one capability to rule them all) is another approach but it doesn't really address it since the ratio "boilerlplate LoC" / "actual get/set commands LoC" is still quite > 1.

- I'd argue that these settings are status. It's just status that is "not important enough to keep it up to date unless required". This is already the case for the consumables capability, which works exactly like this and it already has code (on the MQTT, in my opinion) that is not ideal, and it is also partially true for the carpet mode, since it's not currently stored in the status but it still falls in this category.

> Especially with Viomi, there need to be very good reasons for why Valetudo should consider adopting something for it, because it's not a real vendor with just a single product that is already EOL based on a white-label thing with a hacked-together miio software stack

I particularly disagree on this one since, in my opinion, the main and the most important reason why one would choose install Valetudo is to become independent from manufacturer's decisions to EOL a device and therefore reduce e-waste, maintain ownership and functionality of a device that works perfectly fine.

So IMO the fact whether a device is EOL or not should be the very last driving factor for such decisions, but rather the point here should be how many people are using it and, most importantly, maintaining it going forward. More specifically, I think these decisions should work "the Linux way" (random example: http://lkml.iu.edu/hypermail/linux/kernel/1803.1/06845.html) - nobody's using and maintaining it? Drop it.

Quoting from Telegram:

> Hm is there some config file somewhere which defines the rotation speed? Maybe that could be an option

How is this different than suggesting to put duct tape on the LEDs if they are found annoying?

I think the disagreement comes from the fact that you introduced the capabilities system in order to avoid having two types of functionality — capability — handled by the same component. I think this is a very good design decision.

However, I also think that enforcing the real-world meaning of the word "capability" too strictly to the software design isn't necessarily a good idea and, on the other hand, allowing broader meanings doesn't necessarily prospect catastrophic spaghetti.

All three toggles mentioned during this and other discussions (set_repeat, LEDs enable/disable and carpet turbo) are all toggles share the following characteristics:

- They are bool
- They are not particularly important
- They don't affect Valetudo in any other way since, for what it matters to valetudo, the robot will behave *exactly* the same way, and the difference is exclusively "real world"-facing.
- (Unlike setting the movement mode) they do not require any special handling in other places. In other words, they're not particularly "needy"

In the CS/IT world we see a lot of words being used that go from having extremely loose analogies to the real world to having absolutely none, and that tend to make non-tech-y people laugh. File, directory, zip, parent, child, fork, burn, factory, class, browser, hang, daemon, ham radio, "Wireless Fidelity" and let's also say thank you to Mr. Harold Bluetooth for bringing christianity to Denmark and making mice wireless. /s [relevant](https://www.youtube.com/watch?v=fnWypQz6X2U)

So I fail to see why the word "capability" can't be used to describe the capability of a robot to provide a bunch of potentially (but not necessarily) vendor-specific toggle settings that respect the above 4 requirements.

This especially after considering that the UI code for them would be a `for` loop to add enough items in a list, pulling strings from a dict, and that it IMO also addresses the objection you raised of not wanting to add "unique" features:

> there need to be very good reasons for why Valetudo should consider adopting something for it

 I would understand if they were particularly needy features, but I don't see any problems if the 4 requirements are respected. It is a problem if you need to add a very specific class to tackle that particular unique feature, and that's exactly what I'm trying to avoid by making compromises on the meaning of the word "capability".


Sorry for the WoT :)

Hypfer:
> This is already the case for the consumables capability

Unfortunately yes. I'm not happy with that how it is currently and I'd like to remove that when I find the time to do so since it really doesn't make any sense

> So IMO the fact whether a device is EOL or not should be the very last driving factor for such decisions

To clarify: What I ment by refering to it being EOL was more about the Userbase being relatively small and having no way to really grow which is why I'd like to avoid catering to it, because eventually they will disappear due to malfunctioning devices, upgrades etc.

It's exactly the same with the Roborock v1 where people are unhappy that I'm not adding the ability to manually save/restore maps because they need it for their specific robot and of course they're right, however it's really only relevant for that model which will disappear, because it's old, outdated and pretty much EOL.
Sooner or later the users will switch to another device where these hacks aren't required anymore.

That's one of the key factors that led to Valetudo RE which is to this day the best choice on a V1. The V1 just isn't a particularly great choice of robot

> How is this different than suggesting to put duct tape on the LEDs if they are found annoying?

I think that it's different, because the LED Feature isn't Vendor-Specific. There are at least Roborock and Viomi which behave pretty much the same here so it makes sense to implement that and implement it in a way that uses the same interface on both vendors.

On the other hand, the way Viomi handles two-pass cleanups seems just weird and I'd assume that we'd never see that anywhere else. Therefore, I'd deem it acceptible to not support it in order to minimize deviation from the core set of stuff a robot running Valetudo can do.

> you introduced the capabilities system in order to avoid having two types of functionality — capability — handled by the same component

That's one reason why I've added that. Another reason is that it should be possible to look at the capabilites a robot provides and actually know what it can and cannot do (excluding minor stuff such as only 3 zones or something)

One usage of this are the docs which can easily be autogenerated with no special handling for any capability. Just look at the list and you're done. You know everything that there is to know on that level.

This way, we can easily have that comparison table. We could achieve that by different means, but I like how it is now

You could of course counter some downsides by introducing something like well-known bool toggles but that would still require special handling for something that could easily be done by introducing a capability for it if it's something that will reappear in the future. Every well-known bool toggle is basically something that should be a capability and everything that isn't well-known doesn't fit inside valetudo.


Anyways, appearing on multiple vendors isn't a hard requirement for something to become a capability, because that's not the case the MovementMode capability which would currently be Viomi-only.
However it does make sense to have some kind of choice regarding the way the robot drives and therefore it's safe to assume that this will some day appear elsewhere.

This is not the case with `set_repeat` because clearly something like that would be a parameter like `iterations` which would be part of the actual command one is sending to execute a cleanup.


What I'd really like to avoid is having something like "If you have a robot of vendor A, to do X you need to look there, however if your robot is made by Y, this is actually a switch found there".
That completely misses the point of being a generic abstraction.

Writing this, I had the idea that maybe the Viomi capabilities could transparently remap a start cleanup command with more than one iteration attached to it to starting the cleanup and setting the toggle?
That would make this feature available to Viomi users while maintaining the whole abstraction thing of Valetudo

Hypfer:
I think we need to rethink the MQTT Interface for Zone and Segment cleanup so that it can also accept an Iteration count 🤔

https://github.com/Hypfer/Valetudo/blob/9e2e194bcdacc141fc87e5c49489eb669de9c8a9/lib/entities/core/ValetudoZone.js#L26


..and extend the segment cleanup feature to accept such a thing because currently, only zones do

Depau:
All good points, thanks for reading until the end :)

I'm not completely convinced, however, on the general idea of having absolutely no device-specific quirks or slight differences. I think it's quite a miracle that the currently implemented robots work similarly enough to be able to have this almost perfect abstraction.

However, not everything matches perfectly and I don't think it's always a bad idea to have (a small amount of) quirks somewhere, as long as their potential damage is "constrained", it doesn't cause issues elsewhere, there are enough users and at least one maintainer actively fixing issues with the quirks (or else `git rm`).  If someone wants to take care of whatever is needed for Roborock V1's map and the code in question is confined and easily removable, I think it's *okay* for it to stay even if Roborock V1 users have to follow special instructions to access this workaround. Not *good* but *okay*.

 >> This is already the case for the consumables capability
> Unfortunately yes. I'm not happy with that how it is currently and I'd like to remove that when I find the time to do so since it really doesn't make any sense

Everything in the status or remove and forget? I'd gladly like to see that `setTimeout()` gone. I personally never looked at them and someone reported they were broken for Viomi anyway, though I suspect that removing will cause a general HA graph withdrawal mass hysteria.

I guess there isn't a very compelling reason to remove it right now, as long as it doesn't cause issues elsewhere. It's there, it technically works, it's just very ugly.

> Writing this, I had the idea that maybe the Viomi capabilities could transparently remap a start cleanup command with more than one iteration attached to it to starting the cleanup and setting the toggle?
> That would make this feature available to Viomi users while maintaining the whole abstraction thing of Valetudo

The iterations parameter should do the trick, Viomi may have minimum 1 maximum 2, when `== 2` `set_repeat` is used.

Do the other robots support the iterations for "clean all"? Viomi accepts it for all cleaning commands.

> I think we need to rethink the MQTT Interface for Zone and Segment cleanup so that it can also accept an Iteration count thinking

Payload suggestions?

Currently it's `SEGMENT_ID` or `[SEGMENT_1, ... SEGMENT_N]` for segment cleanup, and `[ZONE_ID_1, ... , N]` for zone cleanup.

As I mentioned in Telegram there's no "homie-way" to handle this sort of payloads and no openHAB way either, so a JSON object could work. 

Technically, the "homie-way" in this case would be a transaction API, you set the topics and send a command to commit or abort the transaction, but it's a bit too complex imo and I don't think HA even allows something like this — OH would need a script.

PR #869, opened 2021-04-28
Easier way of getting the token

bruvv:
## Type of change

Type A:
- [ ] Refactor/Code Cleanup
- [X] Docs
  

# Description (Type A)

as can be found in home assistant: https://www.home-assistant.io/integrations/xiaomi_miio/ this is a 99% succes reate of getting the token really really easy for both iphone and android users.
 

Feature discussion thread:

### DIFF
--- a/docs/_pages/installation/roborock-ota.md
+++ b/docs/_pages/installation/roborock-ota.md
@@ -41,7 +41,25 @@ The irony that this guide suggests using "the cloud" to uncloud your device is n
 
 ## Flashing the firmware image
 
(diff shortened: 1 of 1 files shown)


Hypfer:
Why would one want to use the cloud to get the token when there are non-cloud ways available.

Also, if the device is cloud-provisioned, it will most likely run a firmware that doesn't allow local OTA which would require a factory reset which would invalidate the token

Hypfer:
Also also, why would not using python-miio to discover the token make the whole task easier when you still need python-miio to do the actual flashing?

bruvv:
Good points! I just factory resetted my robot and using the python-miio didn't work for me, using the home assistant way did work for me. Also this is something that is easier for most beginner users.

PR #871, opened 2021-04-28
chore: Rename IntensityPreset(\w*) to PresetSelection\1

Depau:
## Type of change

Type A:
- [x] Refactor/Code Cleanup
- [ ] Docs
  
As discussed in Telegram.

Let's try to merge this and see what breaks first, also to make this easier to review since all that's going on here is renaming stuff.


lib/entities/state/attributes/PresetSelectionStateAttribute.js
```diff
@@ -37,11 +37,10 @@ PresetSelectionStateAttribute.TYPE = Object.freeze({
 });
 
 /**
- *  @typedef {string} PresetSelectionStateAttributeValue
  *  @enum {string}
  *
  */
-PresetSelectionStateAttribute.VALUE = Object.freeze({
+PresetSelectionStateAttribute.INTENSITY = Object.freeze({
```
Depau, line 43: This is so that later we can add, for example, `.MOVEMENT_MODE` and keep it separate from intensity presets.

PR #873, opened 2021-04-28
feat(miio): Unify map presigned URL responses

Depau:
## Type of change

Type A:
- [x] Refactor/Code Cleanup
- [ ] Docs
  

From discussion in #820, tries to unify the map presigned URL handling for all Miio robots into the main MiioRobot.

Of course this needs lots of testing.

### DIFF
--- a/lib/robots/roborock/RoborockValetudoRobot.js
+++ b/lib/robots/roborock/RoborockValetudoRobot.js
@@ -100,6 +100,7 @@ class RoborockValetudoRobot extends MiioValetudoRobot {
                     }
                 });
                 return true;
+            // Roborock does not use the common presigned URL implementation, it requires this specific format.
             case "_sync.gen_tmp_presigned_url":
             case "_sync.gen_presigned_url":
             case "_sync.batch_gen_room_up_url": {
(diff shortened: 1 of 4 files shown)


lib/robots/roborock/RoborockValetudoRobot.js
```diff
@@ -100,20 +100,6 @@ class RoborockValetudoRobot extends MiioValetudoRobot {
                     }
                 });
                 return true;
-            case "_sync.gen_tmp_presigned_url":
```
Hypfer, line 103: We can and must keep roborock as-is, since they don't use the common format for the `_sync.gen_presigned_url` response

Hypfer (changes requested):

lib/robots/roborock/RoborockValetudoRobot.js
Depau: I see. In that case, I can just override it and call the parent method on no match. Or just don't call it.

Depau: I reverted the changes for the roborock robot.

Hypfer:
Tested OK on my dreame d9. Please rebase

Depau:
Done

PR #919, opened 2021-05-12
Publish only changed data to MQTT

definitio:
## Type of change

Type A:
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

## Description

 It's not necessary to resend not changed data because messages have a retain flag.


Hypfer:
While of course appreciated, I have to reject this one, because caching not only introduces a multitude of possible issues but also permanently rests in RAM which may be quite limited.

Since there is no real issue arising from republishing the same data, we can avoid those downsides

PR #961, opened 2021-06-01
Swagger Docs Attempt 4 - Electric Boogaloo

Hypfer:
## Type of change

Type A:
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [x] New core feature

This is a continuation of #892

Todos from https://github.com/Hypfer/Valetudo/pull/892#issuecomment-831835935 still apply


Hypfer:
Pro:
- This works and I don't hate it
- We now have schema validation on config endpoints which is a huge improvement

Con:
- Due to the new dependencies, the valetudo binary has grown by over 6mb


The pkg compression feature could mitigate this, but it's still WIP and not available yet. That's the best option IMO

JS bundlers seem to be frontend only (and terrible) so that also doesn't seem to be a viable option.
Furthermore, I've tried vercel/ncc - the only backend bundler it seems - with underwhelming non-functional results.
Looking at the generated Valetudo binary in a hex editor, I'd say it's definitely doable to have swagger embedded in valetudo without a huge size overhead, since most of that payload just consists of whitespaces.

For now, people with space-limited vacuum robots can just use the upx build which will be still smaller than the pre-swagger non-upx valetudo.

Hypfer:
Man this UI

backend/package.json
```diff
@@ -13,7 +13,10 @@
       "../.git/HEAD",
       "../git/refs/heads/master",
       "../frontend/lib",
-      "../package.json"
+      "../package.json",
+
+      "../node_modules/swagger-ui-dist/swagger-ui.css",
+      "./node_modules/swagger-ui-dist/swagger-ui.css"
```
alexkn, line 19: Why are there 2 swagger-ui.css?

.eslintrc.json
```diff
@@ -79,7 +79,7 @@
         "eol-last": ["error", "always"],
         "jsdoc/check-alignment": "error",
         "jsdoc/check-param-names": "error",
-        "jsdoc/check-tag-names": "error",
+        "jsdoc/check-tag-names": ["error", { "definedTags": ["swagger"] }],
```
alexkn, line 82: I don't think we need that anymore.

Hypfer: Indeed!

backend/package.json
Hypfer: Since the move to the monorepo, I've observed two different `node_modules` locations depending on where the npm install was executed. This way, it will be included from whereever that directory decides to pop up

Hypfer:
Using the WIP pkg compression feature with the brotli algorithm + upx for the base binary, I end up with a 10mb valetudo binary. Without upx, it's ~29mb in size.

There are still some minor bugs in that feature but I'm fairly certain that it will be the solution to this and other issues

PR #967, opened 2021-06-13
New findings in the Viomi map format

halfbakery:
## Type of change

Type A:
- [ ] Refactor/Code Cleanup
- [X] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)
I formalized my findings on the viomi map format. The first commit contains only some minor tweaking. The second part is much more radical. I hope I've found the structure of the viomi map which could help us decoding or gives us the possibility of sanity checking.

Feature discussion thread: https://github.com/Hypfer/Valetudo/discussions/966

### DIFF
--- a/assets/kaitai_structs/viomi_map_file.ksy
+++ b/assets/kaitai_structs/viomi_map_file.ksy
@@ -50,6 +50,117 @@ seq:
     if: feature_flags & 0b000100000000000 != 0
 
 
+enums:
+  pixel_type:
+    0: "outside"
+    1: "scan"
+    10: "room_10"
+    11: "room_11"
+    12: "room_12"
+    13: "room_13"
+    14: "room_14"
+    15: "room_15"
+    16: "room_16"
+    17: "room_17"
+    18: "room_18"
+    19: "room_19"
+    20: "room_20"
+    21: "room_21"
+    22: "room_22"
+    23: "room_23"
+    24: "room_24"
+    25: "room_25"
+    26: "room_26"
+    27: "room_27"
+    28: "room_28"
+    29: "room_29"
+    30: "room_30"
+    31: "room_31"
+    32: "room_32"
+    33: "room_33"
+    34: "room_34"
+    35: "room_35"
+    36: "room_36"
+    37: "room_37"
+    38: "room_38"
+    39: "room_39"
+    40: "room_40"
+    41: "room_41"
+    42: "room_42"
+    43: "room_43"
+    44: "room_44"
+    45: "room_45"
+    46: "room_46"
+    47: "room_47"
+    48: "room_48"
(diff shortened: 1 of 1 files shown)


Depau:
Hi, since you seem to have removed it from some reason, what was called tag_len, containing two into tag and len, and replaced with header section, containing just one is, was actually intended to be a TLV (tag length value) header.

TLV is a simple binary structure for dynamic data that is normally used for this kind of stuff, the tag identifies the content type or format, then you have a length or item count, then the value /content.

And in fact, if you look at your own code, most headers are followed by an unknown u4 (which is likely going to be zero) or a length.

In some instances they do completely ignore this format, just to spice things up, but for the most part it looked like TLV when i checked.

halfbakery:
I know the TLV and it's definitely not a TLV. The purpose of the TLV is to put the binary data into "envelopes" to let the older readers safely skip unknown parts. In this case:
- the tag value is not really a tag, it doesn't identify the inner data. All of the sections have the same value, so it looks like as a magic (to be perfectly precise it's a randomized magic)
- the length value is not in bytes of chunks or blocks, it's a record count (when presents). This length specification doesn't let the reader to safely skip a section without knowing its internal format
- the length value is missing from the following sections:
    - in the map_image section it's 8 bytes, which is `01 00 00 00 01 00 00 00`. This could be (1, 1) as a top-left coordinate of the map. It's just a guess, but the roborock format contains the top-left coordinate too
    - in the charger_position section everything got its place. The missing u4 is the header magic. This section never ever had a length field
    - in the rooms section there's no place for the length either. After the header magic, it just starts with the map list or something

BTW If we accept that the map_id as a magic start each section, we'll get exactly 12 sections for the featureFlag value `0x78ff` which contains 12 non-zero bits.

Depau:
Thanks for the clarification

halfbakery:
BTW I got a map where your parser fails: the `unknown8` part is actually a section header magic (the map_id) + 4 bytes unknown data (`00 00 00 01`). In [this comment](https://github.com/PiotrMachowski/Home-Assistant-custom-components-Xiaomi-Cloud-Map-Extractor/pull/122#issuecomment-854466034) you'll find a map where this section is longer.
```
if (featureFlags & 0b000100000000000) {
            //v6 example: 5b590f5f00000001
            //v7 example: e35a185e00000001
            this.take(8, "unknown8");
            this.parseRooms();
```

Trust me, it's not a TLV :-)

Hypfer:
Can we use those maps as test cases?

halfbakery:
It's not my map. Because it's on a public forum, I think you can use to test your decoder privately. But if you want to put into a public testcase or similar, please consult the owner first. That's why I linked the comment and not just the map alone :-)

PR #969, opened 2021-06-25
Minor Dreame fixes

ccoors:
## Minor Dreame fixes

Type A:
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)
- [x] Add the power save status - alternative: just map to "idle" instead?
- [x] Ignore property changes for audio volume (basically just to remove the "Unhandled property change" log entries)

### DIFF
--- a/backend/lib/entities/state/attributes/StatusStateAttribute.js
+++ b/backend/lib/entities/state/attributes/StatusStateAttribute.js
@@ -37,7 +37,8 @@ StatusStateAttribute.VALUE = Object.freeze({
     CLEANING: "cleaning",
     PAUSED: "paused",
     MANUAL_CONTROL: "manual_control",
-    MOVING: "moving"
+    MOVING: "moving",
+    POWER_SAVE: "power_save"
 });
 
 /**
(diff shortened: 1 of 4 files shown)


backend/lib/entities/state/attributes/StatusStateAttribute.js
```diff
@@ -37,7 +37,8 @@ StatusStateAttribute.VALUE = Object.freeze({
     CLEANING: "cleaning",
     PAUSED: "paused",
     MANUAL_CONTROL: "manual_control",
-    MOVING: "moving"
+    MOVING: "moving",
+    POWER_SAVE: "power_save"
```
Hypfer, line 41: This is not a minor bugfix 👀

ccoors: Debatable.

ccoors: So just map it to idle? That would be fine for me too. Docked is definitely wrong though.

PR #996, opened 2021-07-23
feat: Allow reuse of the bundled nodejs runtime binary

Hypfer:
This is an idea I had where I'm not 100% sure if it really is a good one.

Valetudo uses vercel/pkg to take the Valetudo JS Code and simply glue it onto a statically compiled nodejs runtime binary.
This enables us to install/upgrade Valetudo by copying a single binary which is neat.

What is not neat however is that everything bundled with vercel/pkg will be at least 25MB in size as they all have their own bundled nodejs runtime.
As space is quite limited on the robots, this PR would allow using the nodejs runtime bundled with Valetudo to execute arbitrary other js code. For example, it could be used to add a Telegram Bot Microservice which also runs on the Robot without requiring another 25MB nodejs binary.

The PR adds code that runs before every line of Valetudo code which checks for a ".js" file in the process.argv.
If one is found, it is thrown into `eval()`.
It is also possible to use `--repl` as the first parameter to enter the nodejs REPL interactive shell for quick testing.

The runtime would still use the baked-in v8 options such as the heap size so it's not a direct replacement for a nodejs binary.
Though, I'd assume that it should work well enough for most cases

As Valetudo only uses environment variables and its config file for configuration, there also shouldn't be any collisions by using process.argv

This may or may not be a terrible idea.
Therefore, feedback much appreciated

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -1,37 +1,98 @@
 #!/usr/bin/env node
-const Logger = require("./lib/Logger");
-const process = require("process");
-const Valetudo = require("./lib/Valetudo");
 
-var valetudo = new Valetudo();
+/**
+ * This enables the use of the nodejs runtime packaged with the valetudo binary for other small
+ * js applications on the robot.
+ *
+ * If we detect that
+ *      a) We're packaged
+ *      b1) There's a .js file as the first argument
+ *          OR
+ *      b2) --repl is provided
+ *
+ * we undo all pkg patching and then
+ *      b1) just eval whatever there is.
+ *          OR
+ *      b2) enter REPL
+ */
+if (
+    !(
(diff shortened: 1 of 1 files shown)


Hypfer:
I've figured out why this felt odd:

It doesn't belong into the Valetudo Codebase as it has absolutely nothing to do with Valetudo and everything with the runtime itself.

Therefore, this has been moved into the custom-patched pkg-fetch repo which houses the nodejs base binaries:
https://github.com/Hypfer/pkg-fetch

Adding `--ignore-payload foo.js` will skip Valetudo in its entirety and instead try to execute the foo.js.

To get a repl shell, just create a `repl.js` like so

```js
const repl = require("repl");

repl.start({
    useGlobal: true
});
```

PR #1069, opened 2021-09-05
added help if update-firmware fails with no request was made

wehrstedt:
## Type of change

Type A:
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature


Hypfer:
> The command fails with the message `No request was made` - The update-firmware command cannot determine your IP adress correctly.

This isn't correct. There are actually quite a few things that may cause this message.

Thanks though!

wehrstedt:
> This isn't correct. There are actually quite a few things that may cause this message.

Thats true, but its one issue I struggled with and could potentionally save someone a lot of time.

PR #1081, opened 2021-09-13
MQTT config defaults

ccoors:
## MQTT config defaults

Type A:
- [x] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs

# Description (Type A)

Show the MQTT config defaults in the UI. Also, minor UX fix in Timers.

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -8,6 +8,7 @@ import {
     MapSegmentationActionRequestParameters,
     MapSegmentationProperties,
     MQTTConfiguration,
+    MQTTProperties,
     Point,
     RobotInformation,
     Segment,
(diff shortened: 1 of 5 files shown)


Hypfer:
Are you sure that this doesn't save the defaults to the config?

ccoors:
It does save them. Isn't it a bug when they change?

Hypfer:
They shouldn't change but they also shouldn't be persisted on first save as that would be redundant

ccoors:
Okay. I could set the default value as a placeholder for the text fields?

Hypfer:
What about having a preview topic string somewhere which incorporates these values?

That would also be helpful for people that want to directly interact with MQTT.

Something like

**The MQTT Topic structure will look like this:**

`valetudo/BrownHardToFindSwallow/BatteryStateAttribute/level`

which live updates when editing the textfields maybe? Color coding to match the parameters to its text field could also be an option

ccoors:
That does sound neat. Let's start with the placeholders though, I'll think about that preview thing ;)

image

PR #1092, opened 2021-09-17
Log with SSE and display in new UI

ccoors:
## Log with SSE and display in new UI

Type A:
- [x] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)
## UI
Not great, still just a simple `textarea`. But at least usable and filterable.

image

## SSE
Added `/valetudo/log/content/sse`, which generates an event on every new log line.

### DIFF
--- a/backend/lib/webserver/ValetudoRouter.js
+++ b/backend/lib/webserver/ValetudoRouter.js
@@ -4,6 +4,7 @@ const RateLimit = require("express-rate-limit");
 
 const Logger = require("../Logger");
 const Tools = require("../Tools");
+const {SSEHub, SSEMiddleware} = require("./middlewares/sse");
 
 class ValetudoRouter {
     /**
@@ -28,6 +29,7 @@ class ValetudoRouter {
         });
 
         this.initRoutes();
+        this.initSSE();
     }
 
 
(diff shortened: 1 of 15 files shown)


backend/lib/webserver/ValetudoRouter.js
```diff
@@ -139,6 +141,29 @@ class ValetudoRouter {
         });
     }
 
+    initSSE() {
```
Hypfer, line 144: This will also need a shutdown function similar to what can be seen here:

https://github.com/Hypfer/Valetudo/blob/2d38a929c997e82e733708ccf90896d3e3e6105b/backend/lib/webserver/RobotRouter.js#L139-L143

https://github.com/Hypfer/Valetudo/blob/2d38a929c997e82e733708ccf90896d3e3e6105b/backend/lib/webserver/WebServer.js#L213


as open SSE connections will prevent a successful Valetudo shutdown.
That's the cause of the forced shutdowns we were seeing often when replacing binaries.

Hypfer (changes requested):

PR #1097, opened 2021-09-18
feat(ui): Begin work on robot settings (switches/speaker/map data management)

ccoors:
## Speaker and persistent data control

Type A:
- [x] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Also contains some refactoring to remove code duplications.

image

### DIFF
--- a/backend/lib/robots/mock/MockRobot.js
+++ b/backend/lib/robots/mock/MockRobot.js
@@ -24,6 +24,8 @@ class MockRobot extends ValetudoRobot {
         this.registerCapability(new capabilities.MockWaterUsageControlCapability({robot: this}));
         this.registerCapability(new capabilities.MockSpeakerVolumeControlCapability({robot: this}));
         this.registerCapability(new capabilities.MockSpeakerTestCapability({robot: this}));
+        this.registerCapability(new capabilities.MockKeyLockCapability({robot: this}));
+        this.registerCapability(new capabilities.MockObstacleAvoidanceControlCapability({robot: this}));
         this.registerCapability(new capabilities.MockLocateCapability({robot: this}));
         this.registerCapability(new capabilities.MockWifiConfigurationCapability({robot: this}));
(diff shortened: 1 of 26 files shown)


frontend/src/robot/capabilities/MapControl.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    return (
+        <Card>
+            <CardContent>
+                <Typography variant="h6" gutterBottom>
+                    Persistent Data Configuration
+                </Typography>
+                <Divider/>
+                <Typography variant="body2">
+                    Persistent data is a feature of some robots which allows to save no-go areas and virtual walls. It
```
Hypfer, line 101: This text might be slightly misleading as there are robots (e.g. Dreame vSLAM ones) where you can't disable persistent data. It's always enabled

frontend/src/robot/capabilities/MapControl.tsx
```diff
@@ -0,0 +1,111 @@
...
+                    also allows the robot to drive back to the dock wherever it is and keeps the map from being rotated.
+                </Typography>
+                <PersistentMapControl supported={persistentMapControl}/>
+                {mapReset && <MapResetControl/>}
```
Hypfer, line 105: It might make sense to rethink this and decouple the map reset from the persistent data toggle.

The map reset should probably be relocated to an area with all relevant map management functions while the persistent map toggle should be moved to where all the other toggles (e.g. KeyLock, Carpet Mode, etc) are found

Hypfer (changes requested):

ccoors:
You're right. I probably shouldn't write code that late :/

I can't get back to this for a week or so, if you want to get Timers and Events out to people now might be a suitable time for a release... ? ;)

ccoors:
Okay, updated stuff and added the other toggles:

image

PR #1112, opened 2021-09-28
DND configuration in new frontend

ccoors:
## Type of change

Type A:
- [x] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

image

Also changed the layout to `Masonry` on large screens, because the Grid-spacing was completely off. Looks nice imo :)


Hypfer:
> Looks nice imo :)

I agree as far as the individual components go, however Imo the page is getting a bit too crammed with random stuff that isn't related to each other.
Maybe we could split that up into a bit more thematically-specific chunks/pages or add bigger spacers, some headers or other descriptive text?

PR #1114, opened 2021-09-29
feat(ui): Manual control

ccoors:
## Manual control

Type A:
- [x] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

The "enabled" state has to be tracked somewhere. Yes, tracking it as internal state in the capability implementation is not great, but there is no "get manual control state" command, at least on Roborock (and Dreame?).

image

### DIFF
--- a/backend/lib/core/capabilities/ManualControlCapability.js
+++ b/backend/lib/core/capabilities/ManualControlCapability.js
@@ -38,6 +38,14 @@ class ManualControlCapability extends Capability {
 
     /**
      * @abstract
+     * @return {Promise<boolean>}
+     */
+    async manualControlActive() {
+        throw new NotImplementedError();
+    }
+
+    /**
+     * @abstract
      * @param {ValetudoManualControlMovementCommandType} movementCommand
      * @returns {Promise<void>}
      */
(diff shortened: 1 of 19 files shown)


Hypfer:
Hmm the buttons are a bit too small IMO

For comparison, here's how e.g. roborock does it:



I'm okay with the internal state. I do wonder though if we need some manual override "disable remote control" for edge cases where the browser crashed or something like that

ccoors:
I just noticed that the map does not get updated in manual control mode (at least on Dreame), is that the case on Roborock too? Then I can make similar "full screen" controls.

ccoors:
The Joystick control is impossible unless we know the interval rate. With single button events it doesn't matter if the browser crashes, you can't "hold" a button to send the command multiple times.

Hypfer:
> is that the case on Roborock too? 

I think roborock does allow that but as dreame doesn't, we can just ditch the map

> The Joystick control is impossible

Yep, sorry. I should've picked a screenshot without it

> With single button events it doesn't matter if the browser crashes, you can't "hold" a button to send the command multiple times.

Indeed. Okay, nvm then

ccoors:
Increased the button size and removed the map view. For a round button group we'd need an expert.

image

PR #1122, opened 2021-10-02
feat(docs): List recommended binary for each supported robot

ccoors:
## List recommended binary for each supported robot

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

The amount of questions regarding which binary to use for which robot increased recently. Since the docs are only specific for Dreame, the recommended binary should be listed for each robot.

I basically did some educated guessing, looked at https://dontvacuum.me/robotinfo/ and chose armv7-lowmem for robots with <= 256 MiB RAM. Are there any Roborock aarch64 ones?


Hypfer:
> Are there any Roborock aarch64 ones?

S6 MaxV I think but that thing is irrelevant as it's basically unrootable

Issue #1111, opened 2021-09-27
Viomi. Error while adjusting timezone to UTC


DuBass858:
## Describe the bug

All the time the robot cannot set the correct time zone, therefore the time is not correct.

## To Reproduce

After reboot you can see all the time

### Vacuum Model

ViomiV7
### Valetudo Version

2021.09.1
## Expected behavior

The robot must set the correct time

## Additional context

[2021-07-28T19:24:26.339Z] [DEBUG] local request 5009 set_timezone timed out
[2021-07-28T19:24:26.343Z] [DEBUG] >>> local: {"method":"set_timezone","params":[0],"id":7011}
[2021-07-28T19:24:26.908Z] [DEBUG] local request 6010 set_timezone timed out
[2021-07-28T19:24:26.918Z] [DEBUG] >>> local: {"method":"set_timezone","params":[0],"id":8012}
[2021-09-27T18:40:46.002Z] [INFO] Successfully set the robot time via NTP to 2021-09-27T18:40:46.894Z
[2021-09-27T18:40:48.444Z] [WARN] Token is okay, however we're unable to reach the vacuum { retries: 10, method: 'set_timezone', args: [ 0 ] }
[2021-07-28T19:24:40.509Z] [INFO] Cloud message timed out. Assuming that we're not connected anymore
[2021-07-28T19:24:40.511Z] [ERROR] RobotMQTTHandle: Error while polling the robots state MiioTimeoutError: request timed out:{"method":"get_prop","params":["run_state","mode","err_state","battary_life","box_type","mop_type","s_time","s_area","suction_grade","water_grade","remember_map","has_map","is_mop","has_newmap","timezone"],"id":13}
    at Timeout.timeout [as _onTimeout] (/snapshot/Valetudo/backend/lib/miio/MiioSocket.js:210:32)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7)
[2021-07-28T19:24:44.017Z] [INFO] Cloud connected
[2021-07-28T19:25:12.786Z] [WARN] Error while adjusting timezone to UTC

DuBass858:
Valetudo Version
2021.10.0

[2021-07-28T19:24:45.674Z] [DEBUG] Starting NTP Query for { server: 'pool.ntp.org', port: 123, timeout: 10000 }
[2021-07-28T19:24:45.733Z] [DEBUG] Got Time from NTP Server: 2021-10-02T09:41:53.894Z
[...]

PR #1207, opened 2021-11-08
NTPClient.js workaround for Viomi

### LINKED ISSUE #1156: Date and time resets to wrong date
## Describe the bug
Date and time always is wron on the robot.

## To Reproduce

1. Reboot robot
2. Open logs and find NTP syncronysation message such as "[2021-10-13T16:42:17.002Z] [INFO] Successfully set the robot time via NTP to 2021-10-13T16:42:17.894Z"
3. Wait several minutes or make some actions from ui that adds log mesages(for examlple try to enable manual control)
4. Open log and look at date and time on log messages
Result: date and time is wrong.
Expections: date and time is correct.
### Screenshots

[TimeBug](https://user-images.githubusercontent.com/20066453/137178487-86fe6858-fc38-4c29-a892-b33247585460.png)

### Vacuum Model

Xiaomi Mi Robot Vacuum-Mop P

### Valetudo Version

2021.10.0

## Expected behavior

Date must be correct.

## Additional context
<!-- Add any other context about the problem here. --
!
>

TopicsLP:
Viomi Robots seem to magically change back the time if the time change is greater than 5 minutes.
This will simply write the same time 3 additional times, the time will be offset by 9seconds.

Additionally the date command was adjusted to change the system clock with UTC, as NTP time is UTC the system timezone will not matter.

## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description

I added a stupid workaround when the time offset is greater than 5minutes it will simple set the time 3 additional times (to the same time).
Also I changed the `date -s "<formattedDate>"` to `date -u -s "<formattedDate>"` so the time will be set in UTC.
 
I´m pretty sure following issues should get fixed.
Fixes #1156 #1111

### DIFF
--- a/backend/lib/NTPClient.js
+++ b/backend/lib/NTPClient.js
@@ -67,6 +67,15 @@ class NTPClient {
             this.state = new States.ValetudoNTPClientSyncedState({
                 offset: currentNTPTime.getTime() - preSyncTime.getTime()
             });
+            
+            //Viomi Robots seem to magically change back the time if the time change is greater than 5 minutes.
+            //This will simple write the same time 3 additional times, the time will be offset by 9seconds.
+            if(this.state.offset >= 30000) {
+                Logger.info("NTP offset is greater than 5minutes, define time three additional times");
+                setTimeout(() => {this.setTime(currentNTPTime);},3000);
+                setTimeout(() => {this.setTime(currentNTPTime);},6000);
+                setTimeout(() => {this.setTime(currentNTPTime);},9000);
+            }
 
             Logger.debug("Next NTP sync in " + ntpConfig.interval + " ms");
 
@@ -136,7 +145,7 @@ class NTPClient {
             dateString += date.getSeconds().toString().padStart(2,0);
 
 
-            execSync("date -s \""+dateString+"\"");
+            execSync("date -u -s \""+dateString+"\"");
 
             Logger.info("Successfully set the robot time via NTP to", date);
         } else {


Hypfer:
Nah sorry but that's not mergable.

TopicsLP:
It´s me again,
i would like to get a little bit more feedback,
is my approach to fix the Viomi Date/Time issue not to your liking,
or is it the code that i wrote, after some sleep, i can see that it was not well done and taught trough.

I would have create a draft in my forked version for a rewrite, so far i did not test it.
https://github.com/TopicsLP/Valetudo/commit/de292fa8c00a1bc3ff7679078a32bd076d79d897

Hypfer:
Sure,

the issue with this PR is that it doesn't solve the underlying problem.
It is extremely hacky to just call the date command three times in a row.

This would be already extremely bad if it was only happening on Viomi, however it also applies to all robots - even those which aren't utterly broken pieces of junk like viomis.

The correct solution to this issue is to be found somewhere in the Viomi firmware.
Valetudo can't and most importantly shouldn't do anything about that

Hypfer:
I had an Idea:

Instead of putting this workaround in Valetudo, what about creating a bash script named `/usr/local/bin/date` or similar which then calls the actual `date` binary three times with the `-u` flag as needed here?

Valetudo then wouldn't need to ship any vendor-specific workarounds with the result being the same.
It's likely not _the_ proper solution for viomi but it could be a workaround good enough to stay

PR #1221, opened 2021-11-19
Add Roborock LED controls

supersmile2009:
## Type of change
Type B:
- [X] New capability
- [ ] New core feature

# Description (Type B)
2 core capabilities are added:
- ButtonLightsControlCapability
- StatusLEDControlCapability

Implementations for both capabilities for Roborock are added.

ButtonLights should be supported on all Roborocks (or at least most of them). This capability allows you to enable/disable buttons backlight when robot is fully charged.
StatusLED AFAIK is specific to Roborock S7. I checked the pics and user manuals for most of the recent models, dedicated status LED doesn't seem to be present on them.

Feature discussion thread: https://github.com/Hypfer/Valetudo/issues/553
I started developing this feature when core capability was still present. But then I noticed it got removed recently (I'm assuming because there were no vendor-specific implementations). This implementation is differs from the old one in the sense that it doesn't support LED sub-types. Justification: these aren't the same core features, different commands are sent to the robot. So I'd say these are more like fan speed and water usage: concept is similar, but controlled entities are methods are different.
However, I could re-implement it with sub-types if you prefer it that way.

Note: the description for the button lights used in the UI is probably vendor-specific. It might be a good idea to have that message returned by robot implementation via API. I could return another `description: string` property in the response object along with the `enabled: bool`. What do you think?

### DIFF
--- a/backend/lib/core/capabilities/ButtonLightsControlCapability.js
+++ b/backend/lib/core/capabilities/ButtonLightsControlCapability.js
@@ -1,0 +1,15 @@
+const SimpleToggleCapability = require("./SimpleToggleCapability");
+
+/**
+ * @template {import("../ValetudoRobot")} T
+ * @extends SimpleToggleCapability<T>
+ */
+class ButtonLightsControlCapability extends SimpleToggleCapability {
+    getType() {
+        return ButtonLightsControlCapability.TYPE;
+    }
+}
+
+ButtonLightsControlCapability.TYPE = "ButtonLightsControlCapability";
+
+module.exports = ButtonLightsControlCapability;
--- a/backend/lib/core/capabilities/StatusLEDControlCapability.js
+++ b/backend/lib/core/capabilities/StatusLEDControlCapability.js
@@ -1,0 +1,15 @@
+const SimpleToggleCapability = require("./SimpleToggleCapability");
+
+/**
+ * @template {import("../ValetudoRobot")} T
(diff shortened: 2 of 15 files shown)


Hypfer:
> 2 core capabilities are added

No

supersmile2009:
Neither of these features is actually supported by python-miio at the moment. So people would actually need some coding skills in order to flip those switches, even if we're talking about a one-time affair.

Is there any chance these could be added into valetudo in some shape and form? Maybe not in the shape of core capabilities. A vendor-specific section in the settings maybe, which would work with the capabilities not registered in the core module. I understand it's a whole different topic for discussion. So my question basically is: should I open a discussion thread about a possibility of supporting vendor-specific stuff without touching the core?

Hypfer:
> Neither of these features is actually supported by python-miio at the moment. So people would actually need some coding skills in order to flip those switches, even if we're talking about a one-time affair.

That could be solved by adding documentation though. It's actually quite easy since there's the raw-command feature:
```
 mirobo --token xxx --ip 192.168.xxx.xxx raw-command set_led_status [1]
```

The overall problem with having vendor-specific stuff is that it breaks the whole idea of a vendor-agnostic abstraction.
Also, having things work _well enough_ often leads to them never being done correctly.
What I mean by that is that it could be that a proper solution to a problem would be discovered by thinking about the issue hard enough, but that thinking wouldn't happen as it could also be implemented as a vendor-specific quirk thing

PR #1340, opened 2022-01-07
fix(vendor.dreame): Take customOrder option into account for segment cleaning

### LINKED ISSUE #1147: customOrder Segment Cleanup broken on Z10 Pro
## Describe the bug

The Dreame Z10 Pro does not respect the custom cleanup order for segmented cleanups. 

## To Reproduce

1. Start a segmented cleanup with `customOrder: true` and an unusual segment order via REST/MQTT
2. Observe the robot simply cleaning the "nearest" segment first (or some other metric? But that's what I observed)

#### Segments
image

#### Cleanup command with customOrder
image

#### Result
Still cleans Foyer before Kitchen
image

### Vacuum Model
Dreame Z10 Pro on FW 1056

### Valetudo Version
2021.10.0

## Expected behavior

Either Valetudo should not report `customOrderSupport: true` or the cleanup should be performed in the specified order.

## Additional context

This did work on my D9 on FW 1093. Maybe the command is different or just unsupported? A first look at the code didn't reveal an obvious bug.

**EDIT:** Also happens on Firmware 1099

pkoehlers:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

The `customOrderSupported` flag for Dreame robots currently reports `false` and the `DreameMapSegmentationCapability` ignores the `customOrder` option for segment cleaning.

It always passes an ascending order number along with the segments to clean, which effectively behaves like the roborock with customOrder = true.

This PR makes the dreame implementation behave exactly like the roborock one: when `customOrder` is `false` (or missing), the order number will be the same for all segments, which makes the robot determine the order. 

If `customOrder` is `true`, the old implentation is used and segments will be cleaned in order of appearance.

I'd consider this a bug fix, but of course it **will impact existing integrations, which omit the `customOrder`** but rely on it beeing cleaned in custom order.

Tested on Dreame L10 Pro Firmware 1105 

Related to #1147

### DIFF
--- a/backend/lib/robots/dreame/DreameGen2ValetudoRobot.js
+++ b/backend/lib/robots/dreame/DreameGen2ValetudoRobot.js
@@ -419,7 +419,7 @@ class DreameGen2ValetudoRobot extends DreameValetudoRobot {
             },
             segmentCleaningModeId: 18,
             iterationsSupported: 4,
-            customOrderSupported: false
+            customOrderSupported: true
         }));
 
         this.registerCapability(new capabilities.DreameMapSegmentEditCapability({
--- a/backend/lib/robots/dreame/capabilities/DreameMapSegmentationCapability.js
+++ b/backend/lib/robots/dreame/capabilities/DreameMapSegmentationCapability.js
@@ -67,7 +67,7 @@ class DreameMapSegmentationCapability extends MapSegmentationCapability {
                 typeof options?.iterations === "number" ? options.iterations : 1,
                 fanSpeed,
                 waterGrade,
-                options.customOrder ? i + 1 : 1 //determines the order in which the segments should be cleaned
+                options?.customOrder ? i + 1 : 1 //determines the order in which the segments should be cleaned
             ];
         });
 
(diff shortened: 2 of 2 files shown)


Hypfer:
Is it a good idea though? Why would one want to disable that feature?

pkoehlers:
Personally I'm using it with customOrder only, but thats because I have the sorting implemented in HomeAssistant. I could see somebody just having a simple selection in HA, Homie or wherever and letting the robot pick the shortest route?

Hypfer:
I'm asking because that's not how the offical dreame app is doing it.

Sending this potentially invalid data might do more harm than good

pkoehlers:
Oh I see... I simply did some testing on my robot running Valetudo and it worked fine. 

Maybe it's worth thinking about changing the roborock implementation to use custom order by default to get rid of this roborock only flag then? 

Anyways I'd be fine with closing this PR following your argument.

ccoors:
I am pretty sure this won't fix #1147 - it should not change the current behavior?

pkoehlers:
Sorry, it won't... I edited it in the PR description. This would just make the ordering situation more clear. As I commented on the issue the customOrder works fine for me with and without this PR.

PR #1394, opened 2022-02-01
docs: Hint on patched firmware

sky321:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Short hint for L10 Pro and Z10 Pro delivered with patched firmware.

As there are more and more discussions coming up on the topic of L10 Pro and Z10 Pro with patched firmware I would suggest to include a short hint in the documentation.

Z10 Pro firmware rollback / downgrade #1391

Dreame L10 safe stock firmware versions #1215

### DIFF
--- a/docs/_pages/general/rooting-instructions.md
+++ b/docs/_pages/general/rooting-instructions.md
@@ -27,6 +27,8 @@ It is also recommended to join the [Dreame Robot Vacuum Telegram Usergroup](http
 
 **This was patched in many new firmwares. Do NOT update your robot via the Mi Home app if you want to root.**
 
+Unfortunately, L10 Pro and Z10 Pro robots made after 2021-10 come with a vendor patched recovery firmware version so for those robots you are not able to use this method anymore.
+
 There are other ways to root as well, however this one is very easy and very reliable so use that if you can.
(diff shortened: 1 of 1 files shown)


Hypfer:
> Unfortunately, L10 Pro and Z10 Pro robots made after 2021-10 come with a vendor patched recovery firmware version 

Do they though? As far as I'm aware, all cases of patched robots were caused by either people updating them themselves because they weren't reading the rooting instructions or used robots updated by the previous user that have been sold as new by amazon.

sky321:
Hmmm, probably you are right. This one https://github.com/Hypfer/Valetudo/discussions/1215#discussioncomment-1753246 sound like, brought a new one with new firmware on. But who knows?

I would imaging that Dreame will send out there new robots with newest firmware. But no evidence on my side.

For me it was very helpful to have this sort of information by hand for Roborock devices and their FW, to decide on a robot to buy. 

But if you think it‘s misleading, let’s close the Pull request or park it till there is evidence on this topic.

PR #1467, opened 2022-04-11
chore: Only run the nightly job for the main repository

slovdahl:
## Type of change

Type A:
- [ ] UI Feature
- [x] Refactor/Code Cleanup
- [ ] Docs

# Description (Type A)

The nightly job will be automatically run for all forks at the moment, failing
each night because a release can (obviously) not be published.

### DIFF
--- a/.github/workflows/nightly_build.yml
+++ b/.github/workflows/nightly_build.yml
@@ -14,6 +14,7 @@
       'nightly_build':
         {
           'runs-on': 'ubuntu-latest',
+          'if': 'github.repository_owner == "Hypfer"',
           'steps':
             [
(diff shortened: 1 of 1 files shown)


slovdahl:
Sorry, seems like this didn't work 🤔

bild

Hard to tell from the docs, when all the samples are YAML and not JSON. I used https://github.com/home-assistant/core/blob/dev/.github/workflows/wheels.yml#L19 as inspiration. But I'll test it a bit more thorougly in my fork to see how to fix it. Maybe removing the quotes solves it.

slovdahl:
Fix in #1471.

PR #1468, opened 2022-04-11
chore: Add some default VSCode tasks

slovdahl:
## Type of change

Type A:
- [ ] UI Feature
- [x] Refactor/Code Cleanup
- [ ] Docs

# Description (Type A)

Add some default VSCode "tasks" to make development in VSCode a bit easier.

bild

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,48 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+      {
+        "label": "Build Valetudo frontend",
+        "type": "shell",
+        "command": "npm run build --workspace=frontend",
+        "group": "test",
+        "presentation": {
+          "reveal": "always",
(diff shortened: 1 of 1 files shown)


Hypfer:
Personally, I'm really not a fan of VSCode. That thing feels opaque; hiding something that is undesirable.

I can't really put my finger on why exactly but I don't trust it, I don't like it and I don't want it in my codebase.

slovdahl:
Ok, what about adding `.vscode/` to `.gitignore` in that case? Then anyone that wants to use VSCode can at least keep config files locally.

Hypfer:
Yeah that sounds good. I usually just disable the "automatically add untracked files to git" option so that's why it's not part of the .gitignore yet

`.idea/` also belongs there

slovdahl:
#1470

PR #1485, opened 2022-05-02
fix(vendor.roborock): Multiple fixes for Mijia 1S

004helix:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  

# Description (Type A)

1) Mijia 1S uses an old `app_segment_clean` method which does not support `repeat` and `clean_order_mode` options - just a list of segments
 
2) Fixed path to `device.token` for Mijia 1S

3) Fixed volume levels for Mijia 1S


Hypfer:
> Mijia 1S uses an old app_segment_clean method which does not support repeat and clean_order_mode options - just a list of segments

I wonder if it's just FW `0948` that doesn't support it.


_Usually_ the policy is to only support the latest firmware, however given that you'll likely stay the only person with Valetudo on that robot, I guess it can stay like this.

PR #1494, opened 2022-05-11
Refactor RoborockValetudoRobot.pollMap() and track map nonce in mqtt

004helix:
## Type of change

Type A:
- [ ] UI Feature
- [x] Refactor/Code Cleanup
- [ ] Docs

# Description (Type A)

1. Roborock pollMap() method starts new map refresh loop every time its called -
dummycloud re-connected or robot send ```event.status``` message while in active state.
It causes map refresh loop run at maximum rate (600 msec delay) after cleaning cycle.
```
MiioValetudoRobot.js:

onCloudConnected() {
    Logger.info("Dummycloud connected");
    // start polling the map after a brief delay of 3.5s
    setTimeout(() => {
        return this.pollMap();
    }, 3500);
}
```
```
RoborockValetudoRobot.js:

onIncomingCloudMessage(msg) {
    switch (msg.method) {
        ...
        case "event.status":
            ...
            if (StatusStateAttribute && StatusStateAttribute.isActiveState) {
                this.pollMap();
            }
    ...
}

pollMap() {
   ...
   if (now.getTime() - 600 > this.lastMapPoll.getTime()) {
   ...
       this.sendCloud({"method": "get_map_v1"}).then(res => {
           if (res?.length === 1) {
                ...
                setTimeout(() => {
                    return this.pollMap();
                }, repollSeconds * 1000);
           }
   ...
}
```

2. Mqtt publish new map ~ 3 times after map poll,
nonce tracking in MapNodeMqttHandle.js prevents this.

### DIFF
--- a/backend/lib/mqtt/handles/MapNodeMqttHandle.js
+++ b/backend/lib/mqtt/handles/MapNodeMqttHandle.js
@@ -28,6 +28,7 @@ class MapNodeMqttHandle extends NodeMqttHandle {
         }));
 
         this.robot = options.robot;
+        this.previousMapNonce = null;
 
         this.registerChild(
             new PropertyMqttHandle({
@@ -175,6 +176,12 @@ class MapNodeMqttHandle extends NodeMqttHandle {
         }
         const robot = this.robot;
 
+        if (this.previousMapNonce !== robot.state.map.metaData.nonce) {
+            this.previousMapNonce = robot.state.map.metaData.nonce;
+        } else {
+            return null;
+        }
+
         const promise = new Promise((resolve, reject) => {
             zlib.deflate(JSON.stringify(robot.state.map), (err, buf) => {
                 if (err !== null) {
@@ -217,6 +224,7 @@ class MapNodeMqttHandle extends NodeMqttHandle {
         } catch (err) {
             Logger.error("Error while deflating map data for mqtt publish", err);
         }
+        this.previousMapNonce = null;
         return null;
     }
 
--- a/backend/lib/robots/MiioValetudoRobot.js
+++ b/backend/lib/robots/MiioValetudoRobot.js
@@ -71,7 +71,8 @@ class MiioValetudoRobot extends ValetudoRobot {
         this.fdsUploadInProgress = false;
         this.mapPollingIntervals = {
             default: 60,
-            active: this.config.get("embedded") === true && Tools.IS_LOWMEM_HOST() ? 4 : 2
+            active: this.config.get("embedded") === true && Tools.IS_LOWMEM_HOST() ? 4 : 2,
+            error: 30
         };
         this.expressApp = express();
 
--- a/backend/lib/robots/roborock/RoborockValetudoRobot.js
+++ b/backend/lib/robots/roborock/RoborockValetudoRobot.js
@@ -2,6 +2,7 @@ const capabilities = require("./capabilities");
 const fs = require("fs");
 const Logger = require("../../Logger");
 const RoborockMapParser = require("./RoborockMapParser");
+const Semaphore = require("semaphore");
 const zlib = require("zlib");
 
 const DustBinFullValetudoEvent = require("../../valetudo_events/events/DustBinFullValetudoEvent");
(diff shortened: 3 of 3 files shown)


Hypfer:
Roborock pollMap() method starts new map refresh loop every time its called -
    dummycloud re-connected or robot send event.status message while in active state.
    It causes map refresh loop run at maximum rate (600 msec delay) after cleaning cycle.
    
Why? What exactly causes it to run at the maximum rate? `event.status`?

004helix:
> Why? What exactly causes it to run at the maximum rate? `event.status`? Please elaborate

pollMap() run at the maximum rate.

Suppose the robot is docked:
1. First pollMap() called on valetudo startup here:
https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/MiioValetudoRobot.js#L418
It shedules two timeouts:
- on line 413: https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L413 for 60 seconds (this.mapPollingIntervals.default)
- and on line 424: https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L424 for 5 minutes

2. When the first timeout expires, pollMap() called again (by the timeout function):
On the second run first timeout is expired and second timeout canceled on line 386:
https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L386
Now it schedules again two timeouts:
- on line 413: https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L413 for 60 seconds (this.mapPollingIntervals.default)
- and on line 424: https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L424 for 5 minutes

3. Robot starts moving and the cloud send ```event.status``` with active status. pollMap() called again:
on line 107:
https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L107
Second timeout canceled on line 386:
https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L386
Next it schedules again two timeouts:
- on line 413: https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L413 for 2 seconds (this.mapPollingIntervals.active)
- and on line 424: https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L424 for 5 minutes

**But the previous timeout for 60 seconds was not canceled**. i.e. we have 3 timeouts at this point and only one of them (5 minutes) canceled by each call of pollMap()
Each run pollMap() by ```event.status``` or onCloudConnected() event do not cancel timeout scheduled on line 413:
https://github.com/Hypfer/Valetudo/blob/master/backend/lib/robots/roborock/RoborockValetudoRobot.js#L413

004helix:
Python script to track mqtt map-data and status updates
[map-dump-test.txt](https://github.com/Hypfer/Valetudo/files/8670371/map-dump-test.txt)

Hypfer:
I agree that the map poll logic is flawed, however I'm not happy with the changes in the PR.
I'll try and see how to fix those issues in a different way.

As for the MQTT changes, it's currently working as intended.
There aren't any problems attached to publishing similar looking data to an MQTT topic every few seconds in a local network. But even if there were, making `getMapData` return null even if there is map data wouldn't be a suitable fix.

Hypfer:
https://github.com/Hypfer/Valetudo/commit/a9fa64fcec23639b7367a0d3fa61b84cd07eaebe#diff-86852ef1241fed8dc509816e5718b0988d983b3880c51b8838b7ec1a9b650a97R216-R217

The whole this.lastMapPoll thing is actually pretty old and made very little sense both at the time and also right now.
I just didn't know what else to do back then. 
Thanks for pointing that out :)

004helix:
> I agree that the map poll logic is flawed, however I'm not happy with the changes in the PR.
> I'll try and see how to fix those issues in a different way.

Ok :)

> As for the MQTT changes, it's currently working as intended. There aren't any problems attached to publishing similar looking data to an MQTT topic every few seconds in a local network. But even if there were, making `getMapData` return null even if there is map data wouldn't be a suitable fix.

Published data are the same, not similar (if nonce wasn't changed). MQTT payload are large (10-15 KB). Quite apart from WiFi network usage, you set retain frag in message. This causes the data to be saved to disk on mosquitto server every time map updated and waste of SSD resource.

Right now, I don't undestand why valetudo publish identical map 3 times after map poll.

PR #1502, opened 2022-05-14
docs: added action to goto and zonecleaning

bmhorn:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)
Add missing "action" item to sample payloads of the GoToLocationCapability and the ZoneCleaningCapability in the docs of the MQTT integration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ build/
 
 # VSCode project files
 .vscode/
+.devcontainer/
 
 /backend/lib/res/valetudo.openapi.schema.json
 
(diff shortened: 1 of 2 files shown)


Hypfer:
Thank you for your contribution, however I'm afraid I'll have to close this PR as the value isn't missing as it is currently.
The mqtt code doesn't care about that property. It's only displayed in the UI for completeness sake as the REST interface requires it.

Also, the mqtt docs are autogenerated so this change would be reverted next time that script is executed

bmhorn:
No worries. Thank you for your feedback. ;-)

PR #1524, opened 2022-07-21
docs: dreame D9: add more explicit instructions on resetting Wi-Fi

kruton:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

I had followed partial instructions from another guide that used `wifi_connect` to
configure the access point, so I wasn't sure if resetting Wi-Fi applied to me since
it didn't mention this case. After following the steps to reset Wi-Fi, Valetudo worked well.


Hypfer:
>  you may have already set up Wi-Fi using some other method.

What I don't like about this is that with this change, the guide would suddenly start to assume that people don't follow the guide correctly. While at first glance doing that may just sound innocent and helpful, it also sets a harmful precedent.
To be consistent, now the whole guide would have to assume that the user did not follow it at any opportunity to do so.

As this is not a troubleshooting guide, it should only reflect the one correct way of doing things.
That also avoids a lot of bloaty error handling paragraphs, which usually make people tune out entirely due to the amount of text that they would have to read


Thanks for the input though!

PR #1635, opened 2022-11-04
Clean segments by name.

deafloo:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
- [X] Capability implementation for existing core capability
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Over time, some segment borders blur into each other and I have to slice and combine those segments for crisp room separation. When I do this, the segment name gets deleted and the IDs change. I can rename the segment without a problem, but the IDs get messed up and my MQTT broker needs them. It would be nice to send commands by room name instead of room ID.

### DIFF
--- a/backend/lib/mqtt/capabilities/MapSegmentationCapabilityMqttHandle.js
+++ b/backend/lib/mqtt/capabilities/MapSegmentationCapabilityMqttHandle.js
@@ -51,6 +51,29 @@ class MapSegmentationCapabilityMqttHandle extends CapabilityMqttHandle {
                     }
 
                     await this.capability.executeSegmentAction(segments, requestOptions);
+                } else if (Array.isArray(reqSegments.segment_names) && reqSegments.segment_names.length > 0){
+                    const requestOptions = {};
+
+                    if (typeof reqSegments.iterations === "number") {
+                        requestOptions.iterations = reqSegments.iterations;
+                    }
+
+                    if (reqSegments.customOrder === true) {
+                        requestOptions.customOrder = true;
+                    }
+
+                    const robotSegments = await this.capability.getSegments();
+                    const segments = [];
+
+                    for (const name of reqSegments.segment_names) {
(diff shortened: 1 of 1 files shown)


deafloo:
I have not tested or even compiled that code. Can someone with an expendable vacuum test and confirm it? Mine is kind of 'systemically relevant'

Hypfer:
> It would be nice to send commands by room name instead of room ID.

The way it works now is not like that by accident but on purpose.

Only IDs are unique and thus only IDs can serve as an identifier in an API.
Names could be anything. There could be duplicates or just no names at all.

As the Name to ID mapping is published to MQTT, whatever calls the segment cleanup via MQTT should use that information to fetch the correct ID by name and provide proper feedback to the user if there is no such segment.

> Over time, some segment borders blur into each other and I have to slice and combine those segments for crisp room separation. 

Is it a dreame? If yes then consider updating it to the latest firmware as dreame fixed that degrading segment issue a while ago.

deafloo:
Okay I solved it that way. Just in case someone is interested how to do it, here is my home assistant template based on the **Advanced Services** section from the [Valetudo homepage](https://valetudo.cloud/pages/integrations/home-assistant-integration.html):

```YAML
service: mqtt.publish
data:
  topic: valetudo/robot/MapSegmentationCapability/clean/set
  payload_template: '{"action": "start_segment_action", "segment_ids": [{% for attr in states.sensor.valetudo_vacuumName_map_segments.attributes if is_number(attr)%}{% if state_attr("sensor.valetudo_vacuumName_map_segments", attr) == "segmentName" %}{{ attr }}{% endif %}{% endfor %}], "iterations": 1, "customOrder": true}'
```
Just replace *vacuumName* and *segmentName* accordingly.
Please be aware that it can only trigger one room at a time and that there are no fallback routines if the room does not exist.

@Hypfer Yes, it's a D9 but I am already on the latest FW

PR #1663, opened 2022-12-05
Add implementations for Roborock S7 auto empty dock

mrflo97:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs

# Description (Type A)

I bought an Auto Empty Dock for my Roborock S7 and since it's a non-supported robot, there was no way to control the dock in any manner via Valetudo.
That's why I decided to simply have a look and see if I can add it myself. 
I stumbling upon [this Reddit post](https://www.reddit.com/r/Roborock/comments/oyur9z/s7_auto_empty_dock_home_automation/) which pointed to [this Github comment](https://github.com/rytilahti/python-miio/issues/1107#issuecomment-893867891) that contained the necessary commands to control everything. So all that was left to do was to add the appropriate code to valetudo.

Newly added stuff:

**Feature Implementations:**

- AutoEmptyDockAutoEmptyControlCapability
- AutoEmptyDockManualTriggerCapability

**New Quirk AUTO_EMPTY_LENGTH:**
Lets you control the duration of the auto emptying process when it is triggered by the robot itself, just as in the roborock app.

### DIFF
--- a/backend/lib/robots/roborock/RoborockQuirkFactory.js
+++ b/backend/lib/robots/roborock/RoborockQuirkFactory.js
@@ -14,6 +14,57 @@ class RoborockQuirkFactory {
      */
     getQuirk(id) {
         switch (id) {
+            case RoborockQuirkFactory.KNOWN_QUIRKS.AUTO_EMPTY_LENGTH:
+                return new Quirk({
+                    id: id,
+                    title: "Auto Empty Duration",
+                    description: "Set the dustbin emptying duration when triggered by robot after docking.",
+                    options: ["Smart", "Quick", "Daily", "Max"],
+                    getter: async() => {
+                        const res = await this.robot.sendCommand("get_dust_collection_mode", [], {});
+
+                        if (!(res && res.mode !== undefined)) {
+                            throw new Error(`Received invalid response: ${res}`);
+                        } else {
+                            switch (res.mode) {
+                                case 4:
+                                    return "Max";
+                                case 2:
+                                    return "Daily";
+                                case 1:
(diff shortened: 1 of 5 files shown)


backend/lib/robots/roborock/RoborockQuirkFactory.js
```diff
@@ -14,6 +14,62 @@ class RoborockQuirkFactory {
      */
     getQuirk(id) {
         switch (id) {
+            case RoborockQuirkFactory.KNOWN_QUIRKS.AUTO_EMPTY_LENGTH:
+                return new Quirk({
+                    id: id,
+                    title: "Auto Empty Duration",
+                    description: "Set the dustbin emptying duration when triggered by robot after docking.",
+                    options: ["Smart", "Quick", "Daily", "Strong", "Max"],
```
Hypfer, line 22: As `strong` is omitted in the app according to https://github.com/rytilahti/python-miio/issues/1107#issuecomment-893867891, I'd probably also omit it in Valetudo as there could be a hardware reason for that.
Maybe it can cause trouble due to overloaded motors or similar

Hypfer (changes requested):

backend/lib/robots/roborock/RoborockQuirkFactory.js
mrflo97: Good point. I thought that it wouldn't hurt to add it, but you're right, maybe they had a point in excluding it...

Hypfer (approved):

PR #1668, opened 2022-12-13
supported-robots: Update Dreame W10 info with mapping pass and pin header quirks

USA-RedDragon:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [x] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Right now, the mop pad mapping issue is undocumented and the pin header orientation as well. This should help ease some of the Telegram support burden for this mop.

### DIFF
--- a/docs/_pages/general/supported-robots.md
+++ b/docs/_pages/general/supported-robots.md
@@ -242,6 +242,10 @@ Rooting is fairly easy, only requiring a 3.3v USB UART Adapter and almost no dis
 Because of its port placement, it can be a bit difficult to connect the required cables for rooting.<br/>
 If you're struggling to do that, consider removing the Lid to gain better access to the connector.
 
+While the robot is on its wheels and the front is the far side, the BOOTSEL pin is the top right pin in the housing. 
+
+This vacuum will not run a mapping pass while the mop pads are installed.
+
 Rooting instructions:
 - [UART](https://valetudo.cloud/pages/installation/dreame.html#uart)
 


Hypfer:
While I see the idea behind that, I'd very much like to avoid bloating that overview section with information that is temporary in nature. Both of these things will hopefully become irrelevant with rooting method updates and a newer firmware version.

The reasoning behind that is twofold:
Not only do people simply tune out if faced with too much text, but as soon as information is out there on the public web, it is simply impossible to get rid of it again.
This means that there would always be this online rumor that you have to remove the mops on a W10 for a mapping pass even though it might've been fixed ages ago with a firmware update.


If we'd be sure that the W10 is EOL and would not receive a firmware update, it _may_ have been a different story, although even in that case, the docs would probably the wrong place for that information. Instead, Valetudo should just throw an error message if the mops are attached.


Anyway, while it might not always seem like it, a lot of thought went into the docs including which information to include or omit.

USA-RedDragon:
k

USA-RedDragon:
>  This means that there would always be this online rumor that you have to remove the mops on a W10 for a mapping pass even though it might've been fixed ages ago with a firmware update.

So instead, there's an online rumor that mapping doesn't work. Is that better?

USA-RedDragon:
Also, the orientation of the connector is _not_ temporary information

PR #1691, opened 2023-01-19
Add why it matters to the documentation

dmlemos:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [X] Docs

# Description (Type A)

I was reading an article about the leak of Roomba images, and I thought about the importance of open-source and having projects like these.
I've added a section to the docs on why it matters, with this specific example which for me was quite alarming (but not surprising).

It is a small contribution. Feel free not to use it, or add the text somewhere else.

Thank you for keeping the project alive!

### DIFF
--- a/docs/_pages/general/why-valetudo.md
+++ b/docs/_pages/general/why-valetudo.md
@@ -6,7 +6,6 @@ order: 6
 
 # Why Valetudo?
 
-
 First of all, please do **not** try to convince people to use Valetudo.
 
 We all know how terribly it usually turns out when people try to convince their friends to use linux on their desktop.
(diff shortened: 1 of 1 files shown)


docs/_pages/general/why-valetudo.md
```diff
@@ -15,6 +14,15 @@ Using Valetudo only makes sense if you understand its goals and feel like they a
...
+Even as a paying customer you have no control on how this data gets used. Unfortunately many times you are not even aware this is happening in the first place.
+
+Companies use all sort of techniques for working with data sets, but the data never remains in one place (or it is of no use). It is likely that your private data will eventually be leaked on the internet. For example, [data from Roomba robots ended up on the internet](https://web.archive.org/web/20230117013639/https://www.technologyreview.com/2022/12/19/1065306/roomba-irobot-robot-vacuums-artificial-intelligence-training-data-privacy/). iRobot (the company that makes the _Roomba_) contracted a third-party (ScaleAI) to improve on their data. ScaleAI's then created a project on its subsidiary Remotasks for this work. Data from thousands of customers ended up being seen by workers on this platform, which in turn leaked them on social media.
+
+If you are not comfortable with the "Woman on the toilet", perhaps you wouldn't be comfortable with your robot connecting to the cloud.
```
Landrash, line 24: This last part seems out of context.

Landrash (commented):
Seems like a good addition

Hypfer:
Hmm. While I get where you're coming from, I think it would be better suited for e.g. a personal blog post or similar, as there are a few things that would worry me in context of being part of the docs.

First and foremost, pointing at one specific incident of one specific company and mentioning its name multiple times will likely lead to conflict. If not legal than at least in some other way that would be harmful to the project.

Also, in general, I'd prefer to avoid to link to any third-party sites there, as it might be simply too much for a reader to follow all that. More importantly, those external links may also break in the future. Either by actually getting inaccessible or someone there deciding to change the content to something that could be considered questionable. Mozillas Management doesn't exactly have the best track record in that regard IMO.

Another thought is that this news coverage is one specific event, which - while being a good example for the **why** - doesn't add all too much to what was already said on that docs page IMO.
It might be more convincing for people to have a real example instead of theoretical issues, however I honestly don't want to convince people to use Valetudo. They need to arrive at that conclusion by themselves as they then will be much more pleasant to deal with. If you convince someone to use your stuff, they simply have different expectations.

dmlemos:
Your reasoning makes sense, and I totally understand why this shouldn't be included in the documentation. It didn't seem to fit in the context of a discussion either.

I feel the need for using projects like Valetudo, as well as others that share similar values will only increase with time.

PR #1698, opened 2023-02-01
docs: Add description on how to copy the firmware file with scp to the robot

HeyThereSmileMore:
## Type of change
- [x] Docs

# Description (Type A)
Hi, this pull request would add a section (Option 2) which describes how to copy the firmware file with scp from your device to the robot 

Motivation:
I just installed Valetudo on my S5 and for the upgrade from "factory firmware" → 1886 → 2008 I used "valetudo-helper-miioota". From 2008 to 2034 the "valetudo-helper-miioota" tool does not work for me (I think 2008 is blocking the upgrade, I have read in other forums, that other people are facing the same issue) so I was a bit confused on how to get the firmware file to the robot.

I know, there is a "netcat"  server available → Text from DustBuilder:
_Your job has been added on the blockchain and your firmware image roborock.vacuum.s5_2034_fw.tar.gz is ready.
URL of folder, including the netcat server:_

But for me, it was easier to copy the file with scp, so I thought, maybe, we can add this option as an alternative, which might help other people.

BTW: This is my first contribution, if anything is missing or not as good as it should, please reach out, I'm trying my best 😊

### DIFF
--- a/docs/_pages/usage/upgrading.md
+++ b/docs/_pages/usage/upgrading.md
@@ -27,7 +27,7 @@ If you're using an S5 or V1, the recommended way to upgrade Valetudo is to flash
 
    - Option 2: Download the tar.gz file and copy it with scp from your device to the robot
 
-      - Execute the following to copy the file
+      - Examples of how to copy the file to the robot
 
          ```sh
          #Example 1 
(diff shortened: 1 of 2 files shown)


Hypfer:
Thanks for the PR. I will however have to reject it, as if someone wants to do something different from the guide, they should know how to do it. If they don't know how to do it, they should stick to the guide.

While there certainly are multiple options for a few steps, to keep the guide short and easy to maintain while also making supporting users easier, it's better to stick with one well-defined way of doing things.

PR #1710, opened 2023-02-12
Update Updater.tsx

Staubgeborener:
fix small spelling

## Type of change

Type A:
- [ ] UI Feature
- [x] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)

Fix small spelling in update section


Hypfer:
Uuh that's on purpose.
The previous sentence starts with Also so the next one starts with Also also.
I guess you could call that idiom-ish?

It has an entry on Urban Dictionary from 2010 if that means anything: https://www.urbandictionary.com/define.php?term=also%20also

Staubgeborener:
That's also also okay

PR #1801, opened 2023-04-19
Add map_flag parameter 

shaymdev:
## Type of change

Type A:
- [ ] UI Feature
- [ ] Refactor/Code Cleanup
- [ ] Docs
  
Type B:
- [ ] New capability
- [ ] New core feature

# Description (Type A)
The new Manual Map Segment Trigger quirk fails with no parameter given. I'm adding the same parameter that is found in https://github.com/Hypfer/Valetudo/blob/1d2ad47e9d49eac71f0621d2aa0444acf871acec/backend/lib/robots/roborock/capabilities/RoborockMultiMapMapResetCapability.js#L27 

Fixes # 1800

### DIFF
--- a/backend/lib/robots/roborock/RoborockQuirkFactory.js
+++ b/backend/lib/robots/roborock/RoborockQuirkFactory.js
@@ -215,7 +215,7 @@ class RoborockQuirkFactory {
                     },
                     setter: async (value) => {
                         if (value === "trigger") {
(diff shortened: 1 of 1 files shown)


Hypfer:
Quoting the issue

> It errored with the following error but the map now shows segments in valetudo.
> Error: {'code': -10005, 'message': 'save new map too close'}

Something that throws an error can't be a fix for an issue.
We'd also need to ensure that this doesn't break on the S5

PR #1837, opened 2023-06-08
docs: Add info about swapped Tx/Rx cables and reset button location

jnachtigall:


Hypfer:
Hi, thanks, but I believe that the guide is already clear enough and any additional explanation will actually have the opposite effect as people will tune out if there is too much to read.

jnachtigall:
Hi @Hypfer , well, the point that RX cable is not supposed to be connected to RX is nowhere mentioned in the guide. And it took me a lot of time to try out. I already gave up and just found out accidentically another day in the Telegram group that others had the same problem.

The guide as it is very misleading imho with regards to this. And I really tried to follow word by word... It just mentions the RX and TX ports - and shows labels on a picture - and not the TX cable needs to be connected to RX. 

As another user wrote on Telegram:
> However the Valetudo Dreame doc doesn't mention anything about receive and transmit. It just says you need gnd, rx and tx and you can take a look at the following picture on how to connect them to the robots. So I think for new people it would be nice to just mention that RX -> TX and TX -> RX ;-)

Hypfer:
Every person in this space will always at some point have their first contact with a new technology/tool/etc and gets to learn its quirks. For UART that's usually "TX needs to be connected to RX and vice-versa" among other things.

This is nothing the docs of one specific project could take care of, as it is just part of learning the new technology.
It is also not feasible for the docs of one specific project to provide guidance in the technology used, because where do you start and where do you stop? There are a lot of very different people with very different knowledge sets that might stumble upon this project.

Do you explain UART? Okay. But do you also explain what GND is?
And what about using the Terminal?
Or installing Linux?


I completely get that you had a frustrating experience there, but that's not because of an issue with Valetudo docs.
That's just the general pain associated with learning something new.

Everyone needs to make that RX/TX mistake at least once to learn from that. There's just no way around it. In fact, I'd argue that you'll now be much more likely to remember this quirk of UART connections than if you were just blindly following a tutorial :)

And, as @JayFoxRox said, "even if you happen to do it incorrectly, it will still work after swapping." so the absolute worst that can happen is that someone wastes 30 minutes on learning something first-hand.

jnachtigall:
Ok, no worries. Yes, I learnt something (will have it forgotten within weeks though ;-) I personally have a 20 years Linux and professional software dev background, also fiddling with Arduino some years ago, still a lot of Rasberry Pi - but not so much a hardware background rather software. Hence, I did not know about UART (any more). No worries, thanks for answering and providing Valetudo :-)

PR #1871, opened 2023-08-28
Removed Roborock S8 reference from buying-supported-robots.md

nicololuescher:
## Type of change

- [ x] Docs
- [ ] Refactor/Code Cleanup

# Description

Removed remaining references for Roborock S8 Series, since it is not supported at the moment


Hypfer:
> since it is not supported at the moment

That's not what I said.

PR #1888, opened 2023-10-03
docs: Clarify documentation for Dreame setup.

Johnson145:
## Type of change

- [x] Docs
- [ ] Refactor/Code Cleanup

# Description

Some suggestions for the documentation about the Dreame (Z10 Pro) setup. I think they may save some time for future users.

### DIFF
--- a/docs/_pages/general/supported-robots.md
+++ b/docs/_pages/general/supported-robots.md
@@ -366,6 +366,8 @@ The Dreame Z10 Pro is sold as:
 Rooting is pretty easy, only requiring a 3.3v USB UART Adapter, [the Dreame Breakout PCB](https://github.com/Hypfer/valetudo-dreameadapter) and almost no disassembly.
 All warranty seals stay intact.
 
+Note that the Dreame Breakout PCB is _mandatory_ for newer firmware versions. Only if you're lucky and have an old firmware version along with the required knowledge, the PCB may be optional.
+
 #### Details
 
 **Valetudo Binary**: `aarch64`
--- a/docs/_pages/installation/dreame.md
+++ b/docs/_pages/installation/dreame.md
@@ -15,6 +15,8 @@ To root using this method, you'll need:
 - A 3.3V USB to TTL Serial UART Adapter (like CP2102 or Pl2303)
 - A FAT32 & MBR-formatted USB Stick preferrably with an activity LED
 - Some dupont cables
+- A WiFi-capable computer running Linux (ideally a laptop)
+  - If your Linux computer does not have WiFi, you can combine it with a WiFi-capable Windows computer
 
 Basic linux knowledge and a pry tool will help as well.
 
@@ -36,7 +38,7 @@ Don't worry though as the `install.sh` script included in the firmware built usi
 
 #### Note for advanced users
 
-While the Dreame Breakout PCB greatly simplifies the process, it is not strictly _required_ but
+While the Dreame Breakout PCB greatly simplifies the process, it is not always strictly _required_ but
 just _highly recommended_ to avoid people breaking the connector by jamming in 2.54mm pitch cables or shorting 5V to something.<br/>
 
 <details>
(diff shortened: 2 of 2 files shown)


Hypfer:
Thanks for the input but it's incorrect

> Note that the Dreame Breakout PCB is _mandatory_ for newer firmware versions.

This is not true for example. In fact the docs state the exact opposite:

> While the Dreame Breakout PCB greatly simplifies the process, it is not strictly required but just highly recommended to avoid people breaking the connector by jamming in 2.54mm pitch cables or shorting 5V to something.


> A WiFi-capable computer running Linux (ideally a laptop)

This is mentioned here: https://valetudo.cloud/pages/general/rooting-essentials.html

Johnson145:
Thanks for the quick response along with the given explanations.

Of course, you're absolutely free to reject my suggestions. However, maybe you want to reconsider at least a few changes I made? For me, those notes would have saved some hours. I guess it may be similar for other non-expert readers. As always, the devil is in the details. Let me provide you some background information. So you can decide whether useful or not:

---

> A WiFi-capable computer running Linux (ideally a laptop)
> This is mentioned here: https://valetudo.cloud/pages/general/rooting-essentials.html

Three points about this:
- I read the other page, but wasn't sure about the connection to the final setup page. Maybe it's a good idea to simply _link_ the essentials in the requirement section of the dreame setup?
- The tricky thing is the **WiFi**. Not mentioned on the other link. I was using a _desktop_ computer, because that's my only device running Linux. No WiFi, no way to access the robot's network. I solved this issue by using another Windows laptop having WiFi. Took me some time to figure out whether everything needs to be done on the same computer or not. So I thought that's helpful to point out.
- Laptop is more appropriate than desktop, because arranging the docking station etc. around the USB ports of (my) desktop was pretty tricky. Also not yet mentioned explicitly.

---

> Note that the Dreame Breakout PCB is mandatory for newer firmware versions.
> This is not true for example. In fact the docs state the exact opposite:

You'll know that better than I do. However, exactly because the docs stated that the board is optional, I tried it without and again lost quite some time. Maybe you have a better (correct) adjustment for the docs? Background:

I successfully rooted the same device earlier this year without the board. I re-ordered the same vacuum and read 

> While the Dreame Breakout PCB greatly simplifies the process, it is not strictly required but just highly recommended to avoid people breaking the connector by jamming in 2.54mm pitch cables or shorting 5V to something.

so I thought "Cool, he found a way to make it easier. I already know the manual way, so I'll save some time by doing it again the same way without the board.". So I used the older documentation again which I used earlier this year. That one is still based on the reset button method. When doing so, I ended up getting the 
```
cat: can't open '/etc/os_release': No such file or directory
```
error. In your telegram group someone stated that in this case the firmware is too new and I have to use the PCB. I organized one and it worked. Maybe my actual problem was referring to the old reset button method? Either or, any hint stating something like "Hey, even if you already have done this successfully without the board for the same device, doesn't mean that it still works that way." would have been really helpful to me.


---

> Once you have access to the debug port, arrange the vacuum and its docking station close to your computer. You must be able to wire your computer with the vacuum while it is in its docking station. 

I had to cancel the achieved connection, download etc. and redo everything again, because I couldn't just place the large docking station on top of my desktop computer. Would have been great to know the finally required setup in advance.

---

> Now insert your USB stick. You don't have to put any files or similar on it. It's just needed to trigger the login shell.

I had issues with the board not reacting to the USB stick. As a USB stick is usually required to transfer data, making it explicitly not the case, would have helped me debugging the actual problem faster, too.

---

> If you got space issues, you can unplug the USB stick.

Once again I had to cancel and start again, because my USB stick was "too large" for the heavy docking station of the Dreame Z10 Pro. The setup you described was physically not possible. I thought I might have to use a smaller USB stick, but was brave enough to just remove it at that time and it worked for me. If you think there's no downside of unplugging the stick, it might once again help other people with the large docking station.

Issue #1664, opened 2022-12-05
PSA: Where did all those issues and discussions go?


Hypfer:
Hi and welcome to the annual maintainer burnout prevention issue.
This time, it's purely informational. Please enjoy this wall of text

## "Where are all those old issues and discussions?"

Gone.

## "Yes, I can see that. But why?"

Because their existence was becoming a burden for me.
Let me explain

First, we need to understand that in western culture, there's one very broken mechanic:
As soon as one participant of a discussion shows an emotional reaction, they set a fail flag that takes precedence over everything else. It doesn't matter what they said, what they've done or where they are coming from or even why there's an emotional reaction.

As soon as there is one, they're out.

This mechanic is also being severely abused by a lot of people trying to "win" by doing everything they can to set this flag and then point at it being set. While I assumed that people would stop being like that as soon as they've left high school, the unfortunate truth is that a lot of people never do leave high school. They don't get older. They just age.

Anyway, since I do care deeply about this project, said flag is practically on by default or at least very close to that, meaning that I'm basically set up to constantly fail with those mechanics in place.

It will come as no surprise that continually entering this failure state _because_ you're passionate about something and _because_ you've given something to the world is taxing.

## GitHub flavoured public bugtrackers

With that out of the way, let's talk about GitHub flavoured public bugtrackers.
I don't like them.

Actually that's an understatement. It's often a strong detractor from the fun that is building Valetudo; so strong that it utterly obliterated my motivation to do anything with the project many times now.

### Why is that?

Well, first of all, what do you do when you receive an email which you don't want to answer because it's a stupid question, someone not reading the documentation or someone once again challenging your design decisions?
You delete it. You're done with it.

You don't have to deal with the emotional fallout of the sender noticing that you did that because they don't receive any feedback.
For all they know, you might not have gotten that email at all. Or maybe you did get it but forgot about it.
They'd likely still be unhappy, but not actively hostile as they tend to not perceive you as the aggressor.

What happens when you receive an issue that you don't want to answer because it's a stupid question, someone not reading the documentation or someone once again challenging your design decisions?

#### You delete it?
Ha. That means that now the person on the other end knows that you don't want to deal with that.
This is usually being met with hostility. How dare you silence this person. Someone else could've helped them! If you don't want to help, at least let the community help!!11

#### You keep it but don't answer?
Ha. Now you get to watch effectively dead threads staring at you every time that you look at your inbox to see if there is anything new that needs handling.
[...]

PR #1910, opened 2023-10-25
Add device_class attribute to battery and WiFi sensors for Home Assistant 

mill1000:
## Type of change

- [ ] Docs
- [ ] Refactor/Code Cleanup

# Description

This PR add the `device_class` attribute to the battery and WiFi sensors for Home Assistant. Having a proper `device_class` attribute allows HA to apply some extra styles to entity.

e.g. For battery level, the icon is color-coded depending on state and the device page shows the battery level

**Icon Colored**
image
image

**Battery On Device Page**
image

I didn't notice any obvious changes for the WiFi sensor to report here.

 ## Possible additional changes
 The following additional updates might be possible. 
 * Current Statistics Time -> DeviceClass.DURATION
 * Total Statistics Time -> DeviceClass.DURATION
 * Consumables -> DeviceClass.DURATION 
   * Only if `unit !== stateAttrs.ConsumableStateAttribute.UNITS.PERCENT`
* Status Flag -> Maybe DeviceClass.ENUM?
* Error -> Maybe DeviceClass.ENUM? 
  * Or maybe could be refactored to a binary sensor with BinarySensorDeviceClass.PROBLEM
  
There is also a BinarySensorDeviceClass which can be seen [here]( https://github.com/home-assistant/core/blob/f28c9221e63266a512a3bfef9a66c2dab6ccd96e/homeassistant/components/binary_sensor/__init__.py#L31C5-L116) but I didn't find an applicable classes to the current binary sensors. 

A future improvement could expose if an update is available with the UPDATE class.

### DIFF
--- a/backend/lib/mqtt/capabilities/WifiConfigurationCapabilityMqttHandle.js
+++ b/backend/lib/mqtt/capabilities/WifiConfigurationCapabilityMqttHandle.js
@@ -1,6 +1,7 @@
 const CapabilityMqttHandle = require("./CapabilityMqttHandle");
 const ComponentType = require("../homeassistant/ComponentType");
 const DataType = require("../homie/DataType");
+const DeviceClass = require("../homeassistant/DeviceClass");
 const EntityCategory = require("../homeassistant/EntityCategory");
 const HassAnchor = require("../homeassistant/HassAnchor");
 const InLineHassComponent = require("../homeassistant/components/InLineHassComponent");
@@ -96,7 +97,8 @@ class WifiConfigurationCapabilityMqttHandle extends CapabilityMqttHandle {
                             HassAnchor.REFERENCE.HASS_WIFI_CONFIG_ATTRS
                         ),
                         json_attributes_template: "{{ value_json.attributes | to_json }}",
-                        entity_category: EntityCategory.DIAGNOSTIC
+                        entity_category: EntityCategory.DIAGNOSTIC,
+                        device_class: DeviceClass.SIGNAL_STRENGTH
                     },
                     topics: {
                         "": {
--- a/backend/lib/mqtt/homeassistant/DeviceClass.js
+++ b/backend/lib/mqtt/homeassistant/DeviceClass.js
@@ -1,0 +1,61 @@
+/**
+ * Retrieved from https://github.com/home-assistant/core/blob/8b1cfbc46cc79e676f75dfa4da097a2e47375b6f/homeassistant/components/sensor/const.py#L64-L416 on 2023-10-25.
+ *
+ * See also https://developers.home-assistant.io/docs/core/entity/#generic-properties
+ *
+ * @enum {string}
+ */
+const DeviceClass = Object.freeze({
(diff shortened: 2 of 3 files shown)


Hypfer:
This is not a bug fix as you initially selected. It's a feature for which there is no category in the PR template because I'm not interested in feature PRs. The reason for that is that they're just more trouble then they're worth.

For example, one thought that comes to mind is "how does this affect the minimum required Home Assistant Version?" because there's a schema validation in HAs MQTT feature meaning that this will break hard on older versions, likely making use of Valetudo impossible.
It might still be reasonable to do that but that all depends on how old that minimum required version is.

Did you consider those implications? Probably not.
Which is generally fine, because a lot of projects operate like that but this one doesn't.

> A future improvement could expose if an update is available with the UPDATE class.

No, absolutely not. That will never happen. See also https://github.com/Hypfer/Valetudo/discussions/1890
No further input or opinions required on that btw. Not interested.

mill1000:
Wow. I'm sorry, I didn't know. I understand you're just trying to protect your sanity (e.g. #1664), and I appreciate the work you've done but it saddens me to see such hostility towards potential contributors.

> For example, one thought that comes to mind is "how does this affect the minimum required Home Assistant Version?" because there's a schema validation in HAs MQTT feature meaning that this will break hard on older versions, 
...
> Did you consider those implications? Probably not.

You're right. I hadn't and that's a very valid concern. Personally, I think the beauty of PRs is that we can have that discussion now.

FWIW
The device_class function was added 5 years ago: https://github.com/home-assistant/core/pull/14033
There was a recent related bug fix in Jan 2023: https://github.com/home-assistant/core/pull/85106

Additionally the MQTT discovery documentation (https://www.home-assistant.io/integrations/mqtt#discovery-payload) indicates that unknown keys are ignored to in effort to allow backwards compatibility.

Regards,
Tucker

Hypfer:
> it saddens me to see such hostility towards potential contributors.

I do get that perspective but IMO that is part of that fake story about FOSS, big corpo started telling people to get them to exploit themselves for corporate profits.

The probably a bit harsh truth there is that for this project, contributers 80% of the time are a net negative. They take time to deal with, they usually add stuff to the codebase that then needs to be maintained by yours truly and at the end of the day they're "solving" something there is no shortage of: Coding.

I can do the coding just fine - after all that's why the project even exists.
It doesn't help me if someone does the things that I find enjoyable for me and then leaves me with a bunch of new things that I have to do instead which aren't that. Some mildly while others not at all.

<br/>

I hope you don't take this personally because it's not a personally issue. It's cultural.

Every talking head on the internet tells you that you should contribute because contributing is oh so great, then you do invest the time to actually do that and then suddenly the maintainer tells you to F off.

What a frustrating experience for everyone involved. Everyone but the talking heads of course who continue to bring this nonsense upon random people for their own personal gain.

<br/><br/>

aaanyway, I do very much appreciate that you took the time to put in the effort of researching if there might be any trouble with that change even after having received this rather harsh reply!

PR #2018, opened 2024-02-28
Fix loading images messing up anchors on the supported robot page

CSNWEB:
## Type of change

- [x] Docs
- [ ] Refactor/Code Cleanup

# Description

As discussed on Telegram I tried to fix the issue of anchors pointing to the wrong place on the supported robots page. I am relatively confident, that this fixes the issue, since in different browsers (Safari, Chrome, Firefox) and on throttled internet speeds it prevents the text from reflowing when the images are loaded. However reproducing the original bug of jumping to the wrong place proved to be difficult. I could somewhat consistently trigger it using an simulated iOS device after erasing all caches on it. I tried it several times with these changes and the issue did not occur, so I would say I am 90% certain that it works. 

I did not want to globally overwrite the image styling, so I added a class to all images on the page. The width must be explicitly specified or it doesn't work, thus I set it to 100%. The aspect ratio of the images does not seem to be exactly 4 in all cases, but it seems like its close enough, otherwise it would need to be set for every image individually.

### DIFF
--- a/docs/_pages/general/supported-robots.md
+++ b/docs/_pages/general/supported-robots.md
@@ -112,7 +112,7 @@ Don't assume any compatibility of consumables or other parts as well as rooting 
 
 ### Xiaomi V1<a id="xiaomi_v1"></a>
 
-<img src="./img/robots/xiaomi/xiaomi_v1.jpg"/>
+<img class="aspect-ratio-4" src="./img/robots/xiaomi/xiaomi_v1.jpg"/>
 
 The Xiaomi V1 is made by Roborock. It is sold as:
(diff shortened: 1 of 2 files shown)


Hypfer:
Sorry for leaving this PR in limbo for this long.

I just tried the following

```
/docs/_pages/general/img/robots$ find -iname *.jpg -exec convert {} -gravity center -crop 4:1 +repage {} \;
```

with the result unfortunately being not ideal
image

However, I also saw that WebStorm can do this
image

so I think that would be the way to go.
Thanks though for looking into it! Much appreciated

PR #2038, opened 2024-03-31
Roborock gen4 mcuclock

xuefer:
## Type of change

- [ ] Docs
- [ ] Refactor/Code Cleanup
- [X] Vendor feature

# Description
Sync system time to mcu, which help restoring correct time when booting (and maybe other mcu time relative behavior)

1. in G10s(and pro), vendor busybox date does not allow setting system time, use busybox2 provided by dennis in the rooted system image
2. fake a local.time from Valetudo (as cloud) to AppProxy which then sent to miio_client etc, triggering mcu time sync vendor code. This method works, confirmed by testing on hardware and by reverse engineering of AppProxy/miio_client (etc) binary

### DIFF
--- a/backend/lib/NTPClient.js
+++ b/backend/lib/NTPClient.js
@@ -4,6 +4,7 @@ const ntp = require("@destinationstransfers/ntp");
 const execSync = require("child_process").execSync;
 
 const Logger = require("./Logger");
+const SetMCUTimeCapability = require("./core/capabilities/SetMCUTimeCapability");
 const States = require("./entities/core/ntpClient");
 const Tools = require("./utils/Tools");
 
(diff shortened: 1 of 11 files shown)


Hypfer:
Since `execSync` uses `$PATH`, you can just modify the Valetudo runscript on the affected G10s so that `date` points to the `busybox2` binary.

This way, that robot-specific logic stays robot-specific. No need for any change to Valetudo

xuefer:
> Since `execSync` uses `$PATH`, you can just modify the Valetudo runscript on the affected G10s so that `date` points to the `busybox2` binary.
> 
> This way, that robot-specific logic stays robot-specific. No need for any change to Valetudo

I agree about "date" issue. 

About mcu time, since the mcu uart device is opened by roborock daemons exclusively, no other tool can access mcu. I can't think of any other way to sync mcu time other than triggering daemons' time sync code by "local.time" response.
I'm not sure how vendor do it either: it looks like there's only one time chance, when local proxy (before cloud) connected, it will ask for local.time once which does solve the time drift issue for long uptime. Valetudo do ntp sync periodically is better than the vendor way, for os time at least.

PR #2240, opened 2025-04-04
Fix typo in second fastboot picture

kryma-foulab:
I think this is in scope described in CONTRIBUTING.md

I am mildly annoyed at an inaccuracy and you probably would be too


Hypfer:
Well I don't think this is a "typo". It's just the same image as in the previous step.
I can imagine why this PR started existing. Probably a frustrated hour not understanding why it doesn't work.

I am doubtful that this will actually help much honestly, because if people don't manage to read the text instructing them to pick the right livesuit image, they also won't manage to read the text in the image.

Then again, I can also see that this at least increases the likeliehood that people can spot their mistake on their own so it does provide value.

I would've just wished for more honesty calling it a typo and that it's inaccurate.
You or someone you care about made a mistake. That mistake has been frustrating
Mistakes are human.
The whole topic can be daunting and, evidently, this is one issue that when issues arise at all, it's this issue.

So yeah, worth trying to make harder to do worng, but really let's just be honest in our communication

kryma-foulab:
Fair enough. It was a small wrinkle in the process, mostly on the user's end - calling it a typo was rather strong.

Thanks for onboarding - I won't make that mistake again if I flash another robot.

PR #2330, opened 2025-09-05
feat: Expose current segment location to Home Assistant

Skaronator:
# Summary

- Adds two read-only sensors:
    - current-segment-id
    - current-segment-name
- Values are empty if the robot’s position can’t be mapped to a segment (no map/pose, off-map, or unsegmented).

# Motivation

- Enables simple, dependable segment-based automations without changing core behavior.
- Example automations:
    - Modify water/fan/vacuum mode based on the current room (e.g., increase water on tile; reduce water on parquet).
    - Turn hallway lights to a warning color to avoid slamming the door into the vacuum.
    - Automatically raise a standing desk in the study room when the vacuum enters to prevent cable snags.
    - Close the patio door so the vacuum doesn’t escape.

# Testing

Tested on a Dreame L10s; works in moving mode and cleaning mode.

## Screenshots

image

image

# Why Merge

Exposing the current segment id/name is a narrowly scoped, read‑only addition that doesn’t alter core behavior or broaden the project’s scope; it simply surfaces existing state.

This improves dependability and observability for common segment‑based automations while reducing user guesswork.

The change is backwards‑compatible, low‑risk, and low‑maintenance, aligning with the project’s preference for minimal code that benefits the mainstream use case rather than niche workflows.

### DIFF
--- a/backend/lib/entities/map/ValetudoMap.js
+++ b/backend/lib/entities/map/ValetudoMap.js
@@ -1,6 +1,7 @@
 const crypto = require("crypto");
 
 const MapLayer = require("./MapLayer");
+const PointMapEntity = require("./PointMapEntity");
 const SerializableEntity = require("../SerializableEntity");
 const ValetudoMapSegment = require("../core/ValetudoMapSegment");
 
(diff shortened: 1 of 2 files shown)


Hypfer:
_Especially_ not when it's AI-generated.

Fucking hell

Hypfer:
image


Just for reference, I can see that the name of that repo changed in a way that would imply a hostile fork.

That _instantly_ happening. _Instantly_(!)
... kinda proves the exact judgement that was made here.

You do not chose this path this quickly if your intentions were good and you were just _severely_ misunderstood.
That does not happen.
You will be hurt. You will be angry. But you won't _instantly_ pick that path.


Anyway, meta aside, to quote the contributing.md

> Due to that, pretty much all low-hanging-fruits are already gone and those that still appear 🍐 to be there are either not as trivial and simple as you'd think initially or conscious design decisions that should actually be like that.

There have been prior instances of me pondering that feature that have even been public and could've been found via searching.
I do not expect that of course.

A private "hey btw what about this?" question would have been somewhat fine.
There at least would've already been the communication channel, given that this person received a free PCB from me in 2023.


And this concludes the paper trail so far.

PR #2350, opened 2025-10-01
Fix `DreameAutoEmptyDockAutoEmptyIntervalControlCapabilityV1` unable to handle `interval: off`

jul13579:
> **STOP**
> 
> This project does currently not accept pull requests.<br/>
> Please refrain from opening this PR as it will be closed automatically.
> 
> Thank you for understanding.
> 
> For more information, check out the `Contributing.md` in the root of the repository

Yes, I know. I also read this:

> # Contributing to Valetudo
> 
> **Don't**

Regardless, I still figured you might want to know when something seems to be broken. I'll admit, I'm probably not the best person to point out if something is broken (don't look at the closed PR I created a while ago again, I'm still embarrassed). I'll also accept if my solution isn't what you're looking for, or how you did it yourself. I figured at best I saved you some time fixing something yourself, and at worst you'll just close it.

Having this out of the way: The issue. With Version 2025.10 you merged the `AutoEmptyDockAutoEmptyControlCapability` into the `AutoEmptyDockAutoEmptyIntervalControlCapability`. I understand that some robots seem to model these previously separate capabilities as one capability/state information, however the `Dreame Z10 Pro` I'm running doesn't seem to do so. When trying to configure the dock auto-empty setting to "off" on version 2025.10, I'm just getting a red toast notification stating "Server Error: -1".

I'm unsure if my changes have any negative side-effects on other Dreame bots, and am also not sure if any bots other than Dreame ones show the same issue. I'll obviously also happily stand corrected if the issue I'm describing actually is a non-issue/user error.

Lastly, thank you very much for creating and still maintaining Valetudo. It probably is the single thing in my HomeAssistant-based smart home that I actually have to worry about the least. It just works.

### DIFF
--- a/backend/lib/robots/dreame/capabilities/DreameAutoEmptyDockAutoEmptyIntervalControlCapabilityV1.js
+++ b/backend/lib/robots/dreame/capabilities/DreameAutoEmptyDockAutoEmptyIntervalControlCapabilityV1.js
@@ -15,17 +15,24 @@ class DreameAutoEmptyDockAutoEmptyIntervalControlCapabilityV1 extends AutoEmptyD
         super(options);
 
         this.siid = DreameMiotServices["GEN2"].AUTO_EMPTY_DOCK.SIID;
-        this.piid = DreameMiotServices["GEN2"].AUTO_EMPTY_DOCK.PROPERTIES.INTERVAL.PIID;
+        this.piid_interval = DreameMiotServices["GEN2"].AUTO_EMPTY_DOCK.PROPERTIES.INTERVAL.PIID;
+        this.piid_enabled = DreameMiotServices["GEN2"].AUTO_EMPTY_DOCK.PROPERTIES.AUTO_EMPTY_ENABLED.PIID;
 
         this.helper = new DreameMiotHelper({robot: this.robot});
     }
 
     async getInterval() {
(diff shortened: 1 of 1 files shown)


Hypfer:
> Regardless, I still figured you might want to know when something seems to be broken

Yes, but why as a PR instead of a quick ping in the Telegram group or even just an issue or a discussion.

>  I figured at best I saved you some time fixing something yourself, and at worst you'll just close it.

But there is explicit communication to please not do that.

> Lastly, thank you very much for creating and still maintaining Valetudo. 

You can support me in doing so by adhering to the explicit communication given.
Once you saw the PR template stating "don't do it", a pivot would've been the correct thing to do.

A "thank you" as part of something I explicitly told you not to do just adds insult to injury.

Hypfer:
You see, the thing I'm fighting is that people have "good intentions" but the context in which the intentions are good is not real.

The fictional world in which those intentions are good does not contain me.
It contains a version of what people would _like_ me to be. What people think I _should_ be. 

And that is an issue, because I like being me and I want to be me.
I am not going to be what other people imagine what I might or should be.

That's why there is explicit communication.There is no guessing involved.
There is just active refusal of a reality that feels less pleasant to people, because in that reality, I am actually real and not just a supporting side character to their story.

